PR #187, opened 2026-05-01
fix(nmos/is04): subscription field omitempty bug exposed by real-peer Cerebrum test

by-rune:
## Summary

**Real-peer test against EVS Cerebrum (10.100.0.5:8080) surfaced a codec bug** that loopback round-trips couldn't catch.

The shared \`is04.Subscription\` struct used \`*string + omitempty\` on both \`ReceiverID\` and \`SenderID\`. Per IS-04 v1.3 sender.json, \`receiver_id\` is **required** with type \`["string","null"]\` — the field MUST be on the wire even when null. Go's \`json.Marshal\` drops nil-pointer fields tagged \`omitempty\` entirely, so unrouted Senders went out as \`{"active":false}\` instead of \`{"receiver_id":null,"active":false}\`. Spec violation on every Sender / Receiver registration.

Cerebrum silently added the field back with null in its echo — exactly the false-positive round-trip-clean that hides codec bugs. The user pushed back: "no deviation make sure your enc/decoder is not bugged" — turned out to be ours.

## Fix

- Split \`Subscription\` into two distinct types matching the two distinct subschemas:
  - \`SenderSubscription\` — \`{receiver_id, active}\`
  - \`ReceiverSubscription\` — \`{sender_id, active}\`
- Drop \`omitempty\` on the id field so nil-pointer renders as \`null\` (spec-required).
- Update 3 test call sites.

## Verification

\`\`\`
=== dhs Node API after fix ===
"subscription": {"receiver_id":null,"active":false}      ← Sender, was {"active":false}
"subscription": {"sender_id":null,"active":false}        ← Receiver, was {"active":false}

=== Cerebrum Query API after fix ===
"subscription":{"active":false,"receiver_id":null}       ← round-trips clean
"subscription":{"active":false,"sender_id":null}         ← round-trips clean
\`\`\`

Other Cerebrum-side deviations remain (sender.version blanked, flow.sample_rate dropped, sender.subscription.active flipped) — those are NOT codec bugs on our side; our wire bytes match the spec byte-for-byte. Documenting separately as Cerebrum interop notes.

## Test plan

- [x] \`go build ./...\`
- [x] \`go test ./internal/amwa/...\`
- [x] Live verification: dhs Node registered against EVS Cerebrum 10.100.0.5:8080, both Sender + Receiver subscriptions round-trip correctly.

### DIFF
--- a/internal/amwa/codec/is04/is04_test.go
+++ b/internal/amwa/codec/is04/is04_test.go
@@ -247,7 +247,7 @@ func validSender() Sender {
 		DeviceID:          "3b8be755-08ff-452b-b217-c9151eb21193",
 		ManifestHref:      &href,
 		InterfaceBindings: []string{"eth0"},
-		Subscription:      Subscription{Active: false},
+		Subscription:      SenderSubscription{Active: false},
 	}
 }
 
@@ -292,7 +292,7 @@ func validReceiver() Receiver {
 		InterfaceBindings: []string{"eth0"},
 		Format:            FormatVideo,
 		Caps:              ReceiverCaps{MediaTypes: []string{"video/raw"}},
-		Subscription:      Subscription{Active: false},
+		Subscription:      ReceiverSubscription{Active: false},
 	}
 }
 
--- a/internal/amwa/codec/is04/receiver.go
+++ b/internal/amwa/codec/is04/receiver.go
@@ -16,12 +16,12 @@ import (
 type Receiver struct {
 	ResourceCore
 
-	DeviceID          string         `json:"device_id"`
-	Transport         string         `json:"transport"`
-	InterfaceBindings []string       `json:"interface_bindings"`
-	Format            string         `json:"format"`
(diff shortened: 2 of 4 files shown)


by-rune:
Closing — superseded.

The Subscription split fix (commit `03c1e1b` "fix(nmos/is04): split Subscription into SenderSubscription/ReceiverSubscription, drop omitempty on required-but-nullable id") is **already on `main`**. It got squashed in via PR #189 (`b395f1a`).

Current state on `main`:

- `internal/amwa/codec/is04/sender.go` — `SenderSubscription` + `ReceiverSubscription` are split types, no `omitempty` on the id field.
- `internal/amwa/codec/is04/receiver.go` — uses `ReceiverSubscription`.
- Unit tests + per-version (v1.0/1.1/1.2) tests cover the split.

This branch (`fix/is04-subscription-split`) is now redundant + conflicts with `main`. Closing without merging.

Issue #247, opened 2026-05-04
foundation: identity-string sanitiser (internal/identity/sanitize.go)


by-rune:
## Scope

Add a single sanitiser helper at `internal/identity/sanitize.go` exposing three sink-specific functions for vendor strings coming off the wire.

This is **Phase 0 foundation #3** and the smallest piece of the foundation — implemented first to unblock #245 (DM resolver) and #246 (product.yaml).

## Why three sinks, not one

Vendor strings (`card_name`, `sw_rev`, ...) are written to three different sinks with different escape rules. One generic "clean" function would either over-strip yaml escapes or under-protect filesystem paths.

| Wire input (sample) | Path-segment sink | YAML value sink | JSON meta sink |
| --- | --- | --- | --- |
| `RRS18\x00\x00\x00` (NUL-padded) | `RRS18` | `"RRS18"` | `"RRS18"` |
| `RRS18\r\nFW=1.0` | `RRS18-FW-1.0` | `"RRS18 FW=1.0"` | `"RRS18 FW=1.0"` |
| `../../etc/passwd` | **rejected** (`..` segment, `/`) | `"......etc.passwd"` | verbatim |
| `RRS18; rm -rf /` | `RRS18-rm--rf-` | quoted verbatim | verbatim |
| empty / pure NUL | reject, fingerprint = `unknown` | omit field | omit field |
| 600-byte string | truncate to 64 chars | truncate to 256 | truncate to 256 |
| non-UTF-8 bytes | replace per byte with `?` + compliance event | UTF-8 escape | hex-encode under `_raw` |

## API

```go
package identity

// PathSegment returns a value safe to use as a single filesystem path component.
// Returns ErrUnsafe for inputs that cannot be made safe (empty after strip,
// resolves outside parent, etc).
func PathSegment(s string) (string, error)

// YAMLValue returns a value safe to embed in a yaml string field.
// Always succeeds; returns "" for inputs that should be omitted.
func YAMLValue(s string) string

// JSONString returns a value safe to embed in a JSON string field.
// Non-UTF-8 bytes are hex-encoded under a sibling _raw key by the caller;
// this function returns the printable representation.
func JSONString(s string) string

var (
    ErrUnsafe = errors.New("identity: input cannot be sanitised safely")
)
```

## Path-segment allowlist

`[A-Za-z0-9._-]` only. Reject:
- empty
- leading `.`
- contains `/` or `\`
- equals `..` after strip
[...]

Issue #246, opened 2026-05-04
foundation: product.yaml schema + loader (cross-protocol DM library metadata)


by-rune:
## Scope

`product.yaml` is the cross-protocol metadata file at the root of every DM-library entry: `tests/fixtures/products/<vendor>/<product>/<model>-<sw_rev>/product.yaml`.

This is **Phase 0 foundation #2**. Defines the contract every protocol's persisted schema and identity probe relies on.

## Schema (locked design)

```yaml
product:
  vendor:    "Axon"
  product:   "Synapse"
  model:     "RRS18"
  sw_rev:    "1601"
  hw_rev:    "0100"          # metadata only (not in lookup key today)
  description: "..."

supported_protocols:
  - acp1
  - acp2

identity_probe:
  acp1:
    # ACP1 identity is deterministic (group=1, fixed pids) per spec p.20.
    # No cache needed; this section is documentation-only.
    group: 1
    pids:
      model:    0
      sw_rev:   3
      hw_rev:   4
  acp2:
    # ACP2 has no fixed identity addressing. PIDs resolved at first encounter
    # via alias-scan over getObject catalogue, then persisted here so future
    # connects skip the scan entirely.
    pids:
      model:    1234
      sw_rev:   1235
      hw_rev:   1236
    aliases_seen:
      model:    "Card Type"
      sw_rev:   "Software Version"
      hw_rev:   "Hardware Revision"

walk_metadata:
  walked_at:   "2026-04-22T10:30:15Z"
  walked_by:   "dhs/0.7.0"
  controller:  "Axon Synapse Setup 4.2"
  notes:       "..."
```

## Loader contract

```go
package dmlib

type ProductMeta struct {
    Vendor      string
    Product     string
    Model       string
    SwRev       string
    HwRev       string
    Description string
}

type IdentityProbe struct {
    ACP1 IdentityProbeACP1   `yaml:"acp1,omitempty"`
    ACP2 IdentityProbeACP2   `yaml:"acp2,omitempty"`
    // future: emberplus, probel, ...
}

func LoadProductYAML(path string) (*ProductMeta, *IdentityProbe, error)
func SaveProductYAML(path string, m *ProductMeta, ip *IdentityProbe) error  // atomic
```

## Acceptance

- [ ] Round-trip: write yaml → read back → byte-equal (after deterministic key ordering).
- [ ] Unknown protocol section in `identity_probe`: load succeeds, unknown section preserved on save (forward compat).
[...]

Issue #245, opened 2026-05-04
foundation: DM library runtime resolver (internal/dmlib/)


by-rune:
## Scope

Cross-protocol foundation: a runtime resolver for the Device Model (DM) library at `tests/fixtures/products/<vendor>/<product>/<model>-<sw_rev>/<protocol>/`.

This is **Phase 0 foundation #1**. Unblocks ACP1 `SeedFromDM`, provider DM-library loader, and same-pattern roll-out across all Phase 2 protocols.

## Deliverables

`internal/dmlib/` package, stdlib + `gopkg.in/yaml.v3` only.

```go
type Fingerprint struct {
    Vendor  string  // optional
    Product string  // optional
    Model   string  // required
    SwRev   string  // required
    HwRev   string  // metadata only
    Proto   string  // required: "acp1" | "acp2" | "emberplus" | ...
}

type Schema struct {
    Product       ProductMeta
    IdentityProbe IdentityProbe              // cached PIDs for ACP2 (per-proto)
    Slots         map[int]SlotSchema         // slot -> object catalogue
    Metadata      map[string]any
}

type Resolver interface {
    Resolve(fp Fingerprint) (*Schema, error)            // strict; ErrNotFound if (model, swrev, proto) absent
    LookupAlternate(fp Fingerprint) ([]Fingerprint, error) // same model + proto, different swrev
    Persist(schema *Schema, fp Fingerprint) error       // writes the canonical tree on first walk-on-miss
    Diff(prev, cur *Schema) Diff                         // for `dhs registry diff` CHANGELOG verb
}

func New(rootDir string) Resolver
```

## Path layout (locked by `project_dm_library` memory)

```
tests/fixtures/products/
   axon/
      synapse/
         RRS18-1601/
            product.yaml                   # cross-protocol metadata + identity_probe cache
            acp1/
               slot_1.tree.json
               slot_2.tree.json
               ...
            acp2/
               slot_1.tree.json
               ...
```

- Key = `(model, sw_rev, protocol)`.
- Never deduplicated across `sw_rev`. Each rev gets its own dir.
- One product can support multiple protocols simultaneously (sibling proto dirs).

## Acceptance

- [ ] `Resolve(fp)` returns `*Schema` for an existing fixture and `dmlib.ErrNotFound` for a miss.
- [ ] `Persist(schema, fp)` writes atomically (`.tmp` + rename) and is idempotent.
[...]

PR #267, opened 2026-05-04
feat(identity): sanitiser helpers for path/yaml/json sinks (advances #247)

by-rune:
## Summary

- New `internal/identity/` package with three sink-specific sanitiser helpers for vendor-supplied identity strings off the wire: `PathSegment`, `YAMLValue`, `JSONString`.
- Per-byte invalid-UTF-8 replacement (run-collapse semantics in `strings.ToValidUTF8` are not what the spec table calls for).
- 31 table-driven test cases covering NUL padding, CRLF, path traversal, shell-meta, non-UTF-8, control chars, truncation boundaries.

## Why this exists

Phase 0 foundation, smallest unblocker for the DM-library work in #245 (resolver `Persist`) and #246 (`product.yaml` save). Vendor strings come from the wire raw and need different escape rules per persistence sink — one generic "clean" function would either over-strip yaml escapes or under-protect filesystem paths.

## Decisions

| Sink | Rule |
| --- | --- |
| Path | Strict allowlist `[A-Za-z0-9._-]`. Reject `..` substring, `.` / `..` exact, leading `.` or `-`, empty after strip. Truncate to 64. |
| YAML | CRLF/TAB → space (collapsed), path-separators → `.`, control bytes dropped, non-UTF-8 → `?` per byte, truncate to 256. |
| JSON | Pass-through (`encoding/json` handles escaping). Strip NUL only, non-UTF-8 → `?` per byte, truncate to 256 with UTF-8 boundary safety. |

## Test plan

- [x] `go test ./internal/identity/... -count=1` — 31 cases green.
- [x] `go build ./cmd/dhs` clean.
- [ ] CI race-detector run (gcc not available locally).

## Issue refs

Advances #247. Does not close — per the long-running-branch convention, closure happens on the eventual PR-to-main when all foundation pieces land together.

### DIFF
--- a/internal/identity/sanitize.go
+++ b/internal/identity/sanitize.go
@@ -1,0 +1,185 @@
+// Package identity provides sanitiser helpers for vendor-supplied identity
+// strings (card name, software revision, hardware revision, ...). Wire bytes
+// are untrusted: they may be NUL-padded, CRLF-littered, contain path-traversal
+// sequences, or carry non-UTF-8 bytes. Each persistence sink (filesystem path,
+// YAML value, JSON string) needs different escape rules, so this package
+// exports three sink-specific helpers driven by one shared input pre-clean.
+package identity
+
+import (
+	"errors"
+	"strings"
+	"unicode/utf8"
+)
+
+// ErrUnsafe is returned by PathSegment when the input cannot be sanitised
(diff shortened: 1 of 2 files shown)


by-rune:
Superseded by `8e5bcd1` already on main (same scope, same title). Closing as duplicate. No code impact — main unchanged.

PR #268, opened 2026-05-04
feat(dmlib): runtime resolver for DM library (advances #245)

by-rune:
## Summary

- New `internal/dmlib/` runtime resolver for the per-product schema catalogue at `tests/fixtures/products/<vendor>/<product>/<model>-<sw_rev>/<protocol>/`.
- Lookup key `(Model, SwRev, Proto)`. `HwRev` is metadata only. Schemas are never deduplicated across sw_rev.
- Per-slot snapshots reuse `export.Snapshot` (same JSON shape as `dhs export --format json`), so the loop closes cleanly: live walk → `dmlib.Persist` → schema on disk → `dmlib.Resolve` → seed plugin tree.
- 12 unit tests covering hit/miss/invalid, multi-proto same product, alternate sw_revs, atomic persist, diff (added/removed/changed objects + slots), unsafe-fingerprint rejection.

## Path components flow through the sanitiser

Every component of the on-disk path (`Vendor`, `Product`, `Model`, `SwRev`, `Proto`) is run through `internal/identity/PathSegment` from PR #267, so a malicious `Vendor: "../etc"` rejects with `ErrInvalid` rather than escaping the root.

## What's deferred to #246

- `product.yaml` cross-protocol metadata (Vendor/Product/Description) at the dir root.
- `identity_probe` cache for ACP2 alias-scan results.
- Both lift the `Schema` struct slightly; the public API of `Resolver` stays unchanged.

## Test plan

- [x] `go test ./internal/dmlib/... -count=1` — 12 cases green.
- [x] `go build ./...` clean.
- [x] `go vet ./internal/dmlib/...` clean.
- [x] golangci-lint clean (pre-commit hook).
- [ ] CI race-detector run on the umbrella PR-to-main when foundation bundle lands.

## Stacked on

PR #267 (sanitiser) — `feat/foundation-sanitize` is the base. Merge order: #267 → this.

## Issue refs

Advances #245. Closes on the eventual PR-to-main when the foundation bundle lands together.

### DIFF
--- a/internal/dmlib/dmlib.go
+++ b/internal/dmlib/dmlib.go
@@ -1,0 +1,396 @@
+// Package dmlib is the Device Model library: a runtime resolver for the
+// per-product schema catalogue at
+// tests/fixtures/products/<vendor>/<product>/<model>-<sw_rev>/<protocol>/.
+//
+// One product can carry several protocols (sibling proto dirs); each proto
+// dir holds per-slot tree snapshots in the same JSON shape as
+// `dhs export --format json`. The library is the source-of-truth schema
+// every plugin seeds from on Connect.
+//
+// Lookup key is (Model, SwRev, Proto). HwRev rides along as metadata only.
+// Schemas are NEVER deduplicated across sw_rev — each rev is its own dir.
+//
+// This package is cross-protocol foundation; it has no plugin imports and
+// uses no third-party dependencies.
+package dmlib
+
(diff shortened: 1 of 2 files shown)


by-rune:
Superseded by `57f4bcf` already on main (same scope, same title). Closing as duplicate. No code impact — main unchanged.

PR #273, opened 2026-05-04
feat(acp1): Plugin.GetIdentity (advances #251)

by-rune:
## Summary

- New cross-protocol `protocol.Identifier` interface in `internal/protocol/identity.go` with `CardIdentity` struct (Model / SwRev / HwRev) and `ErrIdentityUnresolved` sentinel.
- ACP1 `Plugin.GetIdentity(ctx, slot)` impl in `internal/acp1/consumer/identity.go` — three deterministic getObject calls at group=1 ids 0/3/4 per spec p.20.
- New `acp1_identity_nak` compliance event fired on Card Label NAK.
- `identityValueAsString` covers every ACP1 numeric type so a non-string identity field still produces a fingerprint component.
- 6 unit tests covering happy path, hard NAK + compliance event, soft NAK partial fingerprint, not-connected, IsZero, compile-time interface assertion.

## Spec mapping

| ACP1 ID | Field | CardIdentity | Required |
| ---: | --- | --- | :-: |
| 0 | Card Label | Model | hard |
| 3 | Software Rev | SwRev | soft |
| 4 | Hardware Rev | HwRev | metadata |

Hard-required = NAK fires compliance event and returns `ErrIdentityUnresolved`. Soft-required = NAK leaves the field empty so DM-library partial-match still has a chance.

## Cross-protocol contract

```go
type Identifier interface {
    GetIdentity(ctx context.Context, slot int) (CardIdentity, error)
}
```

The `Plugin.SeedFromDM` flow (#252) and watch hot-plug enrichment (#254) type-assert `plug.(protocol.Identifier)`. Phase 2 protocols satisfy the same interface — ACP2 with alias-scan, others with their wire-specific probes.

## Test plan

- [x] `go test ./internal/acp1/consumer/... ./internal/protocol/... -count=1` — green (4 new + 2 protocol-level + existing).
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Integration test against Synapse Simulator on `feat/acp1-full`.

## Stacked on

PR #272 (health verb). Merge order: #267 → #268 → #269 → #270 → #271 → #272 → this. **First Phase 1 piece.**

## Issue refs

Advances #251. Closes on the eventual PR-to-main when the ACP1 epic bundle lands together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #272, opened 2026-05-04
feat(cli): consumer health verb (advances #250)

by-rune:
## Summary

- New `dhs consumer <proto> health <host>` verb in `cmd/dhs/cmd_health.go`.
- Prints the 3-layer SessionHealth snapshot (reachable / connected / live) in text or `--json`.
- `--watch` polls at `--interval` (default 5s) and emits a row only when any layer bit flips.
- 4 unit tests covering `healthFlipped` transition logic + text/JSON output formatting.

## Surface

```
$ dhs consumer acp1 health 10.6.239.113
host=10.6.239.113 proto=acp1
  reachable=true
  connected=true
  live=true  (last rx 0.4s ago, threshold 1m30s)
  last_rx=2026-05-04T18:42:38Z
  stale_after=1m30s

$ dhs consumer acp1 health 10.6.239.113 --json
{"host":"10.6.239.113","proto":"acp1","reachable":true,"connected":true,"live":true,...}

$ dhs consumer acp1 health 10.6.239.113 --watch
host=10.6.239.113 proto=acp1
  reachable=true ...     # initial reading always emitted
... (silent while bits stable)
host=10.6.239.113 proto=acp1
  reachable=false ...    # emitted on bit-flip
```

## Plugins not yet wired

The verb type-asserts `plug.(protocol.HealthChecker)`. Plugins that haven't implemented `HealthChecker` yet (every plugin in this PR — implementations land per protocol) return:

```
error: health: protocol "acp1" does not implement HealthChecker yet
```

That's preferred over silently printing zero-valued bits. ACP1's implementation lands under #266; Phase 2 protocols under their respective PRs.

## Test plan

- [x] `go test ./cmd/dhs/... -count=1 -run "TestHealth|TestPrintHealth"` — 4 cases green.
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] CI race-detector run on the umbrella PR-to-main.

## Stacked on

PR #271 (SlotInfo State). Merge order: #267 → #268 → #269 → #270 → #271 → this. **Last Phase 0 piece.**

## Issue refs

Advances #250. Closes on the eventual PR-to-main when the foundation bundle lands together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #260, opened 2026-05-04
acp1(provider): DM-library loader (admin slot load --card resolves via internal/dmlib/)


by-rune:
## Scope

Provider-side DM library loader. When admin says `slot load --slot N --card axon/synapse/RRS18-1601/acp1`, the provider resolves the schema via `internal/dmlib/` and serves it from slot N as if a real card were inserted.

This is **Phase 1 ACP1 #10**. Closes the consumer ↔ provider loop: same DM-library entry drives both sides, so a tree walked from our provider is byte-equal to a tree walked from a real Synapse running the same firmware.

## Loop

```text
1. Consumer walks vendor device:
   GetIdentity -> dmlib.Persist({model, sw_rev, acp1}, schema)
2. Provider loads same DM entry:
   admin slot load --card axon/synapse/RRS18-1601/acp1
   server.tree[N] = dmlib.Resolve(...).Slots[N]
3. Consumer walks our provider:
   tree byte-equal to step 1 (after sanitiser normalisation).
```

## Implementation

- `admin slot load --card <vendor/product/model-rev/proto>`: parse the path, resolve via `dmlib.Resolve`, install into `server.tree[slot]`.
- The loaded schema is the source of truth: getValue returns the values from the schema; setValue mutates and persists in-memory only (provider does NOT write back to the DM library; that's the consumer's role).
- Frame-status announce for the slot fires on `present` after the realistic-timing cascade (#259).

## Acceptance

- [ ] `admin slot load --slot 5 --card axon/synapse/RRS18-1601/acp1` resolves and serves.
- [ ] Walking slot 5 from `dhs consumer acp1 walk` returns the schema's full object catalogue.
- [ ] Provider does NOT write back to the DM library — load is read-only.
- [ ] DM-library miss returns a named error from `admin slot load`, no fallback.
- [ ] Race-detector clean integration test.

## Dependencies

- #245 (resolver), #258 (admin CLI), #259 (state machine).

## Phase

Phase 1 ACP1. Lands on `feat/acp1-full`.

by-rune:
Lands as PR #282 on `feat/acp1-dm-library-loader` -> `feat/acp1-slot-state-machine` (stacked on #273-#281).

- 8 new tests covering card-path parsing edge cases, no-resolver error, bad path, resolver miss, cascade drive, extract drive, fingerprint hand-off.
- admin slot.load + slot.unload verbs are now functional (they were stubs).
[...]

Issue #254, opened 2026-05-04
acp1: hot-plug enrichment in watch (identity probe + seed + optional walk on present transition)


by-rune:
## Scope

When a frame-status announce reports a `no_card -> powerup -> boot -> present` transition for slot N during `dhs consumer acp1 watch`, automatically:

1. Probe identity (`GetIdentity(N)`).
2. Look up the DM library entry for `(model, sw_rev, acp1)`.
3. Seed `p.trees[N]` from the resolved schema.
4. If `--auto-walk-on-plug` set, fire a one-shot walk to refresh values.
5. Emit a `hot-plug` row in the watch output:
   ```
   18:42:14  s5  hot-plug  RRS18@1601  seeded(287 objs)  walk=skipped
   18:42:14  s5  hot-plug  RRS19@2001  no-DM-entry        walk=skipped
   18:43:01  s5  hot-plug  RRS18@1601  seeded(287 objs)  walk=ok(287)
   ```

This is **Phase 1 ACP1 #4**. Together with #252 (SeedFromDM) and #253 (scope flags) makes `watch` produce decoded output throughout a hot-plug cycle without operator intervention.

## State machine

```text
no_card  -> powerup : enrichment=skip (no card to identify)
powerup  -> boot    : enrichment=skip (boot still in progress)
boot     -> present : trigger enrichment block above
present  -> error   : keep tree, emit row, no re-seed
present  -> removed : keep tree as stale, emit row
removed  -> no_card : drop tree, emit row
```

## Acceptance

- [ ] Hot-insert during watch: `boot -> present` triggers identity probe + seed, emits row.
- [ ] Hot-remove during watch: `present -> removed -> no_card` keeps stale tree until removed, then drops it.
- [ ] DM-library miss: enrichment row says `no-DM-entry`, walk falls back per the user-supplied scope flag.
- [ ] `--auto-walk-on-plug` triggers walk after seed completes; without it, walk is skipped.
- [ ] Race-detector clean integration test against `dhs producer acp1` with state-machine sequencing (Phase 1 #9).

## Dependencies

- #251 GetIdentity, #252 SeedFromDM, #253 scope flags.
- #249 SlotInfo State enum (transitions emit the correct state value).

## Phase

Phase 1 ACP1. Lands on `feat/acp1-full`.

by-rune:
[...]

PR #274, opened 2026-05-04
feat(acp1): Plugin.SeedFromDM (advances #252)

by-rune:
## Summary

- New `Plugin.SeedFromDM(slot, snap)` method on the ACP1 consumer in `internal/acp1/consumer/seed.go`.
- Pre-populates `p.trees[slot]` from a DM-library `export.Snapshot` so the announce decoder has type info on the first frame — no blocking walk required at Connect time.
- 7 unit tests covering happy path, value-strip invariant, Meta-override of ACP type, not-connected, nil snapshot, single-entry fallback, missing-slot error.

## Why this exists

The watch verb today shows `raw(N)` and empty labels for several seconds at startup because the announce decoder has no type info until the per-slot walk completes. SeedFromDM closes the gap: with a DM-library entry on disk, the decoder sees full type + label info from the very first announce.

## ACPType reconstruction

`protocol.ValueKind` is widened (`KindInt` covers both Integer i16 and Long i32). On seed-from-disk we don't strictly know which the original wire type was. Strategy:

1. If the snapshot's Object carries `Meta["acp1_type"]` (added by walks at persist time — landing in #260), that value wins.
2. Otherwise fall back to a Kind→ACPType mapping (Integer for KindInt, Byte for KindUint, etc.).

The fallback may pick the wrong wire width for set/get round-trips on KindInt/KindUint. Acceptable degradation — the first GetValue / Walk on a seeded slot replaces the seed with wire-confirmed types.

## Hot-swap

Re-seeding on hot-swap is the caller's responsibility. Watch hot-plug enrichment (#254) detects a fingerprint change via `GetIdentity` and calls `SeedFromDM` again with the new schema.

## Test plan

- [x] `go test ./internal/acp1/consumer/... -count=1 -run TestSeedFromDM` — 7 cases green.
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Integration test against Synapse Simulator on `feat/acp1-full`.

## Stacked on

PR #273 (GetIdentity). Merge order: Phase 0 (#267-#272) → #273 → this.

## Issue refs

Advances #252. Closes on the eventual PR-to-main when the ACP1 epic bundle lands together.

### DIFF
--- a/internal/acp1/consumer/seed.go
+++ b/internal/acp1/consumer/seed.go
@@ -1,0 +1,126 @@
+package acp1
+
+import (
+	"errors"
+	"fmt"
+
+	"dhs/internal/acp1/codec"
+	"dhs/internal/export"
+	"dhs/internal/protocol"
+)
+
+// SeedFromDM pre-populates the in-memory slot-tree cache from a DM-library
+// snapshot so the announce decoder has type info on the first frame —
+// no blocking walk required at Connect time.
+//
+// The seed is value-less by design (schemas don't carry live values).
+// First successful announce / GetValue / Walk on a seeded slot
+// overwrites the type entries with wire-confirmed shapes.
+//
+// Hot-swap is the caller's responsibility (watch hot-plug enrichment
+// re-seeds when GetIdentity returns a different fingerprint).
+//
+// ACPTypes is populated via best-effort mapping from protocol.ValueKind.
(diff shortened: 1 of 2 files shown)


by-rune:
Sub-tracker complete — work landed in main as `38f684c`. Closing the tracker; any residual scope continues on the parent issue if still open.

PR #276, opened 2026-05-05
feat(cli): hot-plug enrichment in watch (advances #254)

by-rune:
## Summary

- Hot-plug enrichment integrated into the watch verb's event-handler loop. On `boot -> present` slot transitions, the enricher probes identity, resolves the DM-library entry, seeds the plugin's tree, optionally walks, and prints a `hot-plug` row.
- New `--dm-library <path>` flag on watch to point at the DM-library root (empty disables the lookup; transitions still print but seed step says `no-resolver`).
- New `seederIface` defined locally in `cmd/dhs/` (not `internal/protocol`) since the protocol tier cannot import `dhs/internal/export`.
- 8 unit tests covering all transition paths.

## State machine (per spec p.20)

| Transition | Action |
| --- | --- |
| `no_card -> powerup` | skip (no card to identify) |
| `powerup -> boot` | skip (firmware booting) |
| `boot -> present` | identity → DM lookup → seed → optional walk |
| `present -> error` | keep tree, emit row |
| `present -> removed` | keep tree as stale, emit row |
| `removed -> no_card` | drop tree, emit row |

## Sample output

```
18:42:14  s5  boot -> present  RRS18@1601  seeded(287 objs)
18:43:01  s5  present -> error  card faulted
18:43:18  s5  present -> removed
18:43:18  s5  removed -> no_card
```

With `--auto-walk-on-plug`:

```
18:42:14  s5  boot -> present  RRS18@1601  seeded(287 objs)
         hot-plug s5   walk=ok(287)
```

## DM-library miss

Identity probe succeeds but no schema on disk:

```
18:42:14  s5  boot -> present  RRS19@9999  no-DM-entry
```

Identity probe NAK (Card Label unreadable):

```
18:42:14  s5  boot -> present  id-unresolved  no-DM-entry
```

## Test plan

- [x] `go test ./cmd/dhs/... -count=1 -run TestHotPlug` — 8 cases green.
- [x] Whole-tree `go test ./...` clean (32 packages).
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Integration test against Synapse Simulator + admin slot load/unload on `feat/acp1-full`.

## Stacked on

PR #275 (watch scope flags). Merge order: Phase 0 (#267-#272) → #273 → #274 → #275 → this.

## Issue refs

Advances #254. Closes on the eventual PR-to-main when the ACP1 epic bundle lands together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #275, opened 2026-05-05
feat(cli): watch discovery-scope flags (advances #253)

by-rune:
## Summary

- New discovery-scope flags on `dhs consumer <proto> watch`: `--slots`, `--no-walk`, `--auto-walk-on-plug`.
- **Default behaviour change**: previously walked every present slot at connect; now walks NOTHING by default. Opt in via `--slot`, `--slots`, or `--auto-walk-on-plug`.
- Walk dispatch extracted into `walkScope` + `parseWalkScope` + `runWalkScope` helpers. 14 test cases covering all flag combinations and parse errors.

## Flag matrix

| Flag | Behaviour |
| --- | --- |
| (no flag, default) | walk nothing; subscribe everything; rely on DM-library seed for type info |
| `--slot N` | walk slot N (legacy single-slot mode) |
| `--slots 1,3,7` | walk listed slots |
| `--slots all` | walk every present slot (legacy walk-all behaviour, now opt-in) |
| `--no-walk` | suppress every on-demand walk; pure announce view |
| `--auto-walk-on-plug` | walk on `no_card -> present` transition (logic lands in #254) |

`--no-walk` and `--slots` are mutually exclusive (parse error).

## Why the default change

User locked the new behaviour in the `project_acp1_epic` design discussion: "sometimes we don't want to discover all the cards on a frame". The walk-all default caused walk-storm latency at startup and surprised operators who only wanted frame-status. New default is the explicit opt-in.

`--auto-walk-on-plug` is plumbed but the actual hot-plug detection (boot → present transition triggers a walk on the new card) lands in #254.

## Test plan

- [x] `go test ./cmd/dhs/... -count=1` — all green including 14 new scope tests.
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Integration test against Synapse Simulator on `feat/acp1-full`.

## Stacked on

PR #274 (SeedFromDM). Merge order: Phase 0 (#267-#272) → #273 → #274 → this.

## Issue refs

Advances #253. Closes on the eventual PR-to-main when the ACP1 epic bundle lands together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #256, opened 2026-05-04
acp1(provider): AN2/TCP (Mode C) listener


by-rune:
## Scope

Add the ACP1 AN2 / TCP (Mode C) listener to `dhs producer acp1` on port 2072.

This is **Phase 1 ACP1 #6**. Final transport mode for ACP1 (after Mode A UDP shipped, Mode B TCP-direct in #255).

## Wire model

AN2 is a TCP framing layer over which ACP1 is carried. The `internal/transport/an2` framer already exists for ACP2; reuse it here.

## Implementation

- Listener on TCP :2072.
- Each accepted connection: AN2 framer reads length-prefixed frames, body is an ACP1 PDU dispatched to the same handler set as Mode A and Mode B.
- AN2 protocol-events negotiation per spec.
- One goroutine per session; bounded send-channel; ctx-cancellable.

## Acceptance

- [ ] `dhs producer acp1 serve --transport an2 --port 2072` listens and accepts.
- [ ] `dhs producer acp1 serve --transport all` serves UDP + TCP-direct + AN2 simultaneously.
- [ ] Synapse Setup connects via AN2 and walks a slot successfully (integration test).
- [ ] AN2 protocol-events negotiation completes before the first announce is emitted.
- [ ] DoS guard, frame max bound, race-detector clean — all per #255 acceptance.

## Dependencies

- #255 (Mode B) — same handler set, lands first to validate the dispatcher refactor.

## Phase

Phase 1 ACP1. Lands on `feat/acp1-full`.

by-rune:
Lands as PR #278 on `feat/acp1-an2-listener` -> `feat/acp1-tcp-direct` (stacked on #273-#277).

- 7 new tests green covering GetVersion, ACP1 round-trip, events-enable gating, GetSlotInfo, registry per-IP cap, enableProtos toggle.
- ACP1 mode coverage now matches the spec: Mode A UDP shipped, Mode B TCP in #277, Mode C AN2 here.
- Per-session enableProtos flag — announces only fan to sessions that called EnableProtocolEvents([1]).
- UDP-side broadcastAnnounce bridge fans to AN2 sessions in addition to TCP.
- AN2 framer imported from internal/acp2/codec (stdlib-only); future refactor moves to internal/transport/an2.

by-rune:
[...]

Issue #255, opened 2026-05-04
acp1(provider): TCP direct (Mode B) listener


by-rune:
## Scope

Add the ACP1 TCP direct (Mode B) listener to `dhs producer acp1`. Today the provider serves UDP direct (Mode A) only; spec-correct emulation requires every transport mode the protocol defines.

This is **Phase 1 ACP1 #5**.

## Spec context

ACP1 defines three transport modes (spec §X):

| Mode | Transport | Port | Frame |
| --- | --- | --- | --- |
| A | UDP direct | 2071 | raw ACP1 frame |
| B | TCP direct | 2071 | MLEN-prefix + ACP1 frame |
| C | AN2 / TCP | 2072 | AN2 framing + ACP1 frame |

Mode A is shipped today. This PR adds Mode B; Mode C is filed separately as #256.

## Implementation

- Reuse the existing announce + reply path (the codec is identical; only framing differs).
- New listener accepts TCP, reads MLEN-prefixed frames (bounded to spec max), dispatches to the same handler set.
- One goroutine per session; bounded send-channel; ctx-cancellable.

## Acceptance

- [ ] `dhs producer acp1 serve --transport tcp --port 2071` listens and accepts.
- [ ] `dhs producer acp1 serve --transport all` (default) serves UDP + TCP simultaneously.
- [ ] Real Synapse Setup connects via TCP and walks a slot successfully (integration test, env-gated).
- [ ] DoS guard: refuse > 32 concurrent sessions per source IP (default; flag-overridable later).
- [ ] Frame max bounded per spec; oversize frames -> spec-defined error + compliance event.
- [ ] Race-detector clean unit test for framer.

## Dependencies

None (codec already exists).

## Phase

Phase 1 ACP1. Lands on `feat/acp1-full`.

by-rune:
Lands as PR #277 on `feat/acp1-tcp-direct` -> `feat/acp1-hotplug-enrichment` (stacked on #273-#276).

- 8 new test cases green covering round-trip, oversized MLEN session-close, per-IP cap, registry remove, broadcast drop-on-full, IP extraction, multi-session announce.
- Concurrency: one acceptor + per-session reader/writer + 256-deep bounded send channel. 32 sessions/IP DoS guard.
- Bridge: UDP-side broadcastAnnounce also fans out to TCP sessions.
- CLI: `--transport udp|tcp|all` + `--tcp-port` on `dhs producer acp1 serve`.

[...]

PR #277, opened 2026-05-05
feat(acp1): provider TCP direct (Mode B) listener (advances #255)

by-rune:
## Summary

- New `ServeTCP(ctx, addr)` method on the ACP1 provider in `internal/acp1/provider/tcp_server.go`.
- TCP listener with MLEN-prefixed framing per spec p.10. One acceptor goroutine; per session one reader + one writer goroutine fed from a 256-deep bounded channel.
- DoS guard: 32 concurrent sessions per source IP. Frame max bound at 1024 bytes (well above the 141-byte spec ceiling).
- Bridge: UDP-side `broadcastAnnounce` now also fans out to every live TCP session so a set on UDP still reaches TCP-only consumers.
- CLI: `dhs producer acp1 serve --transport [udp|tcp|all] --tcp-port <p>`. `all` runs both transports concurrently.
- 8 unit + integration tests covering happy path, oversized MLEN, per-IP cap, registry remove, drop-on-full, IP extraction, multi-session announce.

## Spec mapping

| ACP1 mode | Transport | Port | Frame |
| --- | --- | --- | --- |
| A | UDP direct | 2071 | raw ACP1 frame |
| **B (this PR)** | **TCP direct** | **2071** | **MLEN(u32 BE) + ACP1 frame** |
| C | AN2/TCP | 2072 | AN2 framing + ACP1 frame |

Mode C lands separately in #256.

## Concurrency model

- Acceptor: one goroutine, blocks on `Accept`.
- Reader: per-session; handles each frame inline (handleRequest is fast).
- Writer: per-session; drains the bounded send channel.
- Registry: serialises adds/removes; broadcast iterates with read lock.

Slow consumer cannot back-pressure the announce path or other sessions — broadcast drops on full channel + emits a warn log.

## Test plan

- [x] `go test ./internal/acp1/provider/... -count=1 -run "TestServeTCP|TestTCPSession|TestRemoteIP"` — 8 cases green.
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Live test against Synapse Setup over TCP (integration on `feat/acp1-full`).

## Stacked on

PR #276 (hot-plug enrichment). Merge order: Phase 0 (#267-#272) → #273 → #274 → #275 → #276 → this.

## Issue refs

Advances #255. Closes on the eventual PR-to-main when the ACP1 epic bundle lands together.

### DIFF
--- a/cmd/dhs/cmd_producer.go
+++ b/cmd/dhs/cmd_producer.go
@@ -50,6 +50,8 @@ func runProducer(ctx context.Context, protoName string, args []string) error {
 		announceObj   = fs.Int("announce-demo-obj", 18, "acp2: obj-id for --announce-demo target (must be Number+Float)")
 		announceEvery = fs.Duration("announce-demo-interval", 2*time.Second, "--announce-demo tick interval")
 		metricsAddr   = fs.String("metrics-addr", "", "if set (e.g. ':9100'), serve Prometheus /metrics + Go/process collectors on this address")
+		transport     = fs.String("transport", "udp", "acp1 only: udp (default, Mode A), tcp (Mode B), or all (UDP+TCP simultaneously). Other protocols ignore this flag.")
+		tcpPort       = fs.Int("tcp-port", 0, "acp1 only: TCP listen port for --transport tcp/all (0 = same as --port)")
 	)
 	if err := fs.Parse(args); err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


by-rune:
Sub-tracker complete — work landed in main as `40bb35c`. Closing the tracker; any residual scope continues on the parent issue if still open.

PR #278, opened 2026-05-05
feat(acp1): provider AN2/TCP (Mode C) listener (advances #256)

by-rune:
## Summary

- New `ServeAN2(ctx, addr)` method on the ACP1 provider in `internal/acp1/provider/an2_server.go`. Listens on TCP :2072 by default; reads AN2-framed messages and routes by proto field.
- AN2-internal (proto=0): minimal `GetVersion` / `GetDeviceInfo` / `GetSlotInfo` / `EnableProtocolEvents` handlers.
- ACP1 (proto=1): payload is decoded and dispatched through the existing `handleRequest`; reply wrapped in AN2 frame.
- Per-session `enableProtos` flag — announces only fan to sessions that called `EnableProtocolEvents([1])`.
- Bridge: UDP-side `broadcastAnnounce` also fans to AN2 sessions in addition to TCP.
- CLI: `dhs producer acp1 serve --transport [udp|tcp|an2|all] --an2-port`. `all` now runs all three transports.
- 7 unit + integration tests covering `GetVersion`, ACP1 round-trip, events-enable gating across sessions, `GetSlotInfo`, registry per-IP cap, `enableProtos` toggle.

## Spec mapping

| ACP1 mode | Transport | Port | Frame | Status |
| --- | --- | --- | --- | --- |
| A | UDP direct | 2071 | raw ACP1 | shipped |
| B | TCP direct | 2071 | MLEN + ACP1 | #277 |
| **C (this PR)** | **AN2/TCP** | **2072** | **AN2(8B) + ACP1** | **here** |

ACP1 mode-coverage now matches the spec.

## Bridge architecture

```
UDP setX -> broadcastAnnounce -> {bcast UDP, broadcastTCPAnnounce, broadcastAN2Announce}
TCP setX -> handleRequest -> reply on TCP + reg.broadcast (TCP fan-out + bridge)
AN2 setX -> handleRequest -> reply on AN2 + an2Registry.broadcastACP1
```

## Cross-protocol AN2 framer note

AN2 is the transport layer for both ACP1 Mode C and ACP2. Today the framer lives in `internal/acp2/codec/`; ACP1's provider imports it as `an2 "dhs/internal/acp2/codec"`. The codec is stdlib-only and lift-to-own-repo ready, but the cross-protocol dependency is awkward. A future refactor moves AN2 to `internal/transport/an2/` so neither protocol bears the dependency. Out of scope for this PR.

## Test plan

[...]

Issue #258, opened 2026-05-04
acp1(provider): admin CLI surface over local socket (slot load/unload/state, value set/get, reload)


by-rune:
## Scope

Provider admin control plane via local socket / Windows named pipe. CLI subcommands talk over it; REST defers to future `dhs-srv` per ADR-0002.

This is **Phase 1 ACP1 #8**. The pattern lifts to every other protocol's provider in Phase 2.

## Architecture

| Layer | Detail |
| --- | --- |
| Transport | `unix:///var/run/dhs-acp1-<name>.sock` on Linux/macOS; `\\.\pipe\dhs-acp1-<name>` on Windows |
| Default name | `dhs-acp1` (one instance); per-instance via `--name <id>` |
| Wire format | length-prefixed JSON request / response (codec stays internal; not part of public CLI contract) |
| Auth | local-only by default; remote admin opt-in deferred to dhs-srv |

## CLI surface (consumer side of the socket)

```text
dhs producer acp1 serve  --tree T.json --transport all --bind 10.6.239.200:2071 --name rackA
dhs producer acp1 admin  slot load   --slot 1 --card axon/synapse/RRS18-1601/acp1
dhs producer acp1 admin  slot unload --slot 1
dhs producer acp1 admin  slot state  --slot 1 --state error
dhs producer acp1 admin  value set   --slot 1 --group control --id 4 --value Off
dhs producer acp1 admin  value get   --slot 1 --group control --id 4
dhs producer acp1 admin  reload
```

`--name` selects which instance to talk to (multi-instance support).

## Implementation notes

- Admin RPC server runs as a goroutine inside `dhs producer ... serve`, not a separate process.
- Socket / named-pipe path is logged at startup so the user can verify connectivity.
- Every admin call mutates the live tree, fires the same announce path as wire-driven changes (so consumers watching see operator-driven changes too).
- `slot load --card <vendor/product/model-rev/proto>` resolves via the DM library (#245).
- `reload` re-reads the served tree without dropping sessions.

## Acceptance

- [ ] All listed admin verbs work end-to-end against a running provider.
- [ ] Multiple `--name` instances co-exist without socket-path collision.
- [ ] Admin call from a different process succeeds within 50 ms.
- [ ] Consumer's `watch` sees announces fired by every admin mutation.
- [ ] Race-detector clean integration test for serve + admin + watch round-trip.

## Dependencies

- #245 (DM library resolver) for `slot load --card`.
[...]

Issue #257, opened 2026-05-04
acp1(provider): Broadcasts-gated announce path (slot 0 control.4 = Off => silence)


by-rune:
## Scope

Implement the ACP1 Broadcasts gate per spec: slot 0 / control / id 4 (Broadcasts) is an enum that controls whether the device emits spontaneous announces. `On` (default) = emit; `Off` = silence every spontaneous announce until toggled back.

This is **Phase 1 ACP1 #7**. First compliance touchpoint mirroring real Synapse behaviour.

## Spec rule

- The Broadcasts field is a real ACP1 enum on every Synapse frame.
- When `Off`, the device DOES NOT emit frame-status or value-change announces.
- Replies to active requests stay on regardless. Only spontaneous announces are gated.
- Toggling resumes immediately; no flush of buffered events.

## Implementation

The enum value already lives in the served tree. The announce path checks the live tree value before broadcasting. No separate flag, no shadow state — the spec field IS the gate.

```go
// announce_path.go (sketch)
func (s *server) emitAnnounce(frame Frame) {
    if !s.tree.GetEnum(0, "control", 4).Equals("On") {
        return // gated: silent
    }
    s.broadcast(frame)
}
```

## Acceptance

- [ ] With Broadcasts=On (default served tree): `dhs consumer acp1 watch` sees frame-status + value-change announces.
- [ ] After `dhs producer acp1 admin value set --slot 0 --group control --id 4 --value Off`: `dhs consumer acp1 watch` sees zero announces.
- [ ] Active request/reply still works during Off (consumer get/set/walk completes).
- [ ] Toggling back to On resumes announces immediately on the next state change.
- [ ] Race-detector clean unit test for the gate check.

## Dependencies

- #258 (admin CLI) — to flip the field from the test harness.

## Phase

Phase 1 ACP1. Lands on `feat/acp1-full`.

by-rune:
Lands as PR #279 on `feat/acp1-broadcasts-gate` -> `feat/acp1-an2-listener` (stacked on #273-#278).

- 13 new tests covering no-gate-object permissive default, enum On/Off resolution, bare-numeric truth table (8 cases), nil-param fallback, end-to-end broadcastAnnounce gated and flowing.
- Centralised the announce path: TCP and AN2 wire handlers now also route through broadcastAnnounce so the gate, UDP broadcast, and fan-out bridges all apply uniformly.
[...]

PR #279, opened 2026-05-05
feat(acp1): provider Broadcasts gate (advances #257)

by-rune:
## Summary

- New `tree.broadcastsEnabled()` method consults `(slot=0, group=control, id=4)` per spec p.20. The served tree value IS the gate; no shadow flag.
- `broadcastAnnounce` checks the gate before fan-out and bails silently on Off. Replies to active requests stay on regardless — only spontaneous announces are gated.
- Centralised: `handleAN2ACP1Frame` and `serveTCPSession` previously fanned announces directly to their session registries, bypassing the gate + UDP broadcast. Both now route through `broadcastAnnounce`.
- `broadcastAnnounce` no longer requires a UDP socket: standalone TCP/AN2 providers (no UDP listener) emit through the bridge fan-outs without warning.
- 13 unit tests covering no-gate permissive default, enum On/Off, bare-numeric truth table, nil-param fallback, end-to-end gated/flowing broadcasts.

## Resolution order for the gate value

| Tree state at (slot=0, control, id=4) | Result |
| --- | --- |
| Entry absent | `enabled=true` (permissive default) |
| Enum with `EnumMap` | label match: `"On"` (case-insensitive) ⇒ true |
| Bare numeric | `0` ⇒ false; anything else ⇒ true |
| Nil `param` | `enabled=true` |

Legacy trees without the gate object keep their current behaviour. Trees that explicitly set `Off` get full silence. Trees that flip mid-stream resume immediately on the next `Set`.

## Why centralise the announce path

Three paths previously bypassed `broadcastAnnounce`:

1. UDP wire (already went through it) ✓
2. TCP wire (`serveTCPSession`) — fanned to TCP registry directly ✗
3. AN2 wire (`handleAN2ACP1Frame`) — fanned to AN2 registry directly ✗

After this PR all three share one entry point that:
- Honours the Broadcasts gate
- Sends UDP limited-broadcast (when configured)
- Bridges to TCP and AN2 sessions

So a Set arriving on AN2 reaches both UDP and TCP consumers, all subject to the same gate.

## Test plan

- [x] `go test ./internal/acp1/provider/... -count=1` — green (13 new + all existing).
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Live test: flip Broadcasts via admin CLI (#258) and verify silence on consumer side.

## Stacked on

PR #278 (AN2/TCP listener). Merge order: Phase 0 (#267-#272) → #273 → #274 → #275 → #276 → #277 → #278 → this.

## Issue refs

Advances #257. Closes on the eventual PR-to-main when the ACP1 epic bundle lands together.

### DIFF
--- a/internal/acp1/provider/an2_server.go
+++ b/internal/acp1/provider/an2_server.go
@@ -218,21 +218,9 @@ func (s *server) handleAN2ACP1Frame(f *an2.AN2Frame, sess *an2Session) {
 		}
 	}
 	if ann != nil {
-		body, err := ann.Encode()
-		if err == nil {
-			out := &an2.AN2Frame{
-				Proto:   an2.AN2ProtoACP1,
-				Slot:    f.Slot,
-				MTID:    0,
-				Type:    an2.AN2TypeData,
-				Payload: body,
-			}
-			if b, err := an2.EncodeAN2Frame(out); err == nil {
-				if s.an2Registry != nil {
-					s.an2Registry.broadcastACP1(b)
-				}
-			}
-		}
+		// Route through broadcastAnnounce for the Broadcasts gate
+		// (#257) plus the UDP+TCP bridges.
+		s.broadcastAnnounce(ann)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


by-rune:
Sub-tracker complete — work landed in main as `863d23c`. Closing the tracker; any residual scope continues on the parent issue if still open.

PR #284, opened 2026-05-05
feat(acp1): provider fuzz verb (advances #262)

by-rune:
## Summary

- New `dhs producer acp1 fuzz [flags]` verb that drives random valid value-changes against the writable objects in the served tree.
- `RunFuzz` method on the provider performs target collection (writable + type-fuzzable + filter match), random synthesis bounded by metadata, applyMutation + announce per tick.
- Reproducible RNG via `--seed`, rate-limited by `--rate`, time-bounded by `--duration`, scope filters via `--slot --group --id`, boundary bias via `--include-edges`.
- 7 unit tests pinning the contract.

## Flag map (matches issue spec)

| Flag | Behaviour |
| --- | --- |
| `--seed N` | reproducible (same seed → same sequence) |
| `--rate N` | events per second (default 1) |
| `--duration D` | time-bounded; default 0 = until Ctrl-C |
| `--slot N` / `--group G` / `--id N` | filter scope |
| `--include-edges` | bias every 4th cycle to a min/max boundary |

## Bounds-respect by type

| Type | Range | Wire width |
| --- | --- | --- |
| int / long / byte | uniform `[min, max]` | 2 / 4 / 1 byte BE |
| float | uniform `[min, max]` | 4 byte BE IEEE-754 |
| enum | uniform across `EnumMap` | 1 byte |
| ipaddr | uniform 4-byte | 4 byte |
| string | not in this PR (needs phrase pool) | — |

Read-only entries are filtered out at target collection. Filter that matches no writable objects fails fast with a clear error.

## Test plan

- [x] `go test ./internal/acp1/provider/... -count=1 -run "TestRunFuzz|TestSynthInteger"` — 7 cases green.
- [x] Whole-tree provider tests still green.
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Live: `dhs producer acp1 fuzz --tree T.json --rate 100 --duration 60s` against a connected `dhs consumer acp1 watch`.

## Stacked on

PR #283 (fan-out audit). Merge order: Phase 0 (#267-#272) → #273 → #274 → #275 → #276 → #277 → #278 → #279 → #280 → #281 → #282 → #283 → this.

## Issue refs

Advances #262. Closes on the eventual PR-to-main when the ACP1 epic bundle lands together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #283, opened 2026-05-05
test(acp1): provider fan-out audit (advances #261, subsumes #137)

by-rune:
## Summary

Three new audit tests pin the multi-session fan-out properties the issue calls out. No production code changes — the existing implementation already satisfies the issue acceptance criteria; this PR adds the test harness that proves it.

## What's covered

| Test | Pins |
| --- | --- |
| `TestFanout_NoHeadOfLineBlocking` | 31 consumers + 1 trigger; one consumer is designated slow (stops reading); fan-out delivers ≥50% of 50 announces to every healthy consumer |
| `TestFanout_NoGoroutineLeak` | open 16 sessions, close them, GC, verify goroutine count returns to baseline +/- 4 |
| `TestFanout_BroadcastIsConstantTime` | broadcastTCPAnnounce iterates the session map once and pushes the same bytes to each — no per-session encode |

## Why "no production change"

Each session having its own bounded send channel + drop-on-full broadcast were both already in the design (#255 TCP, #256 AN2). This PR is the audit harness the issue requested: pin the contract so future regressions surface in CI.

## Stress + leak details

- 31 + 1 = 32 = `tcpMaxSessionsPerIP` (all loopback dials share 127.0.0.1). Sized to fit under the cap.
- Goroutine leak threshold +/- 4 leaves slack for Go runtime housekeeping (sweepers, GC workers).
- Broadcast contract test uses unbuffered channel inspection — direct pointer-equivalence on the payload.

## Test plan

- [x] `go test ./internal/acp1/provider/... -count=1 -run TestFanout -timeout 30s` — 3 cases green.
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Race-detector run on the umbrella PR-to-main (gcc not available locally).

## Stacked on

PR #282 (DM library loader). Merge order: Phase 0 (#267-#272) → #273 → #274 → #275 → #276 → #277 → #278 → #279 → #280 → #281 → #282 → this.

## Issue refs

Advances #261 (subsumes the older #137 fan-out audit ticket). Closes on the eventual PR-to-main when the ACP1 epic bundle lands together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #282, opened 2026-05-05
feat(acp1): provider DM-library loader (advances #260)

by-rune:
## Summary

- `server.SlotLoad(ctx, slot, cardPath)` resolves a DM-library entry via the configured `dmlib.Resolver` and drives `CascadeInsert` (#259).
- `server.SlotUnload(slot)` drives `CascadeExtract`.
- `admin slot.load` / `slot.unload` verbs now route through these methods (they were stubs).
- Producer CLI: `--dm-library <path>` configures the resolver.
- 8 unit tests covering parse-path edge cases, no-resolver error, bad path, resolver miss, cascade drive, extract drive, fingerprint hand-off.

## Card path syntax

```
vendor/product/model-rev/proto
```

| Path | Vendor | Product | Model | SwRev | Proto |
| --- | --- | --- | --- | --- | --- |
| `axon/synapse/RRS18-1601/acp1` | axon | synapse | RRS18 | 1601 | acp1 |
| `lawo/vsm/GIO-12-2000/acp2` | lawo | vsm | GIO-12 | 2000 | acp2 |

The model-rev split uses the LAST hyphen so hyphenated models survive.

## What's deliberately NOT in this PR

Schema-into-tree replacement. The served `--tree` stays the source of truth for object values; `slot.load` only models the hot-plug cascade so consumers see the transition.

The canonical `<-> []protocol.Object` converter that would actually swap tree contents is a follow-up. Integration tests against Synapse Setup / VSM Studio / Cerebrum will tell us whether the cascade alone is sufficient or whether we need the converter.

## Test plan

- [x] `go test ./internal/acp1/provider/... -count=1 -run "TestParseCardPath|TestSlotLoad|TestSlotUnload"` — 8 cases green.
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Live test: `dhs producer acp1 serve --dm-library tests/fixtures/products/` + `dhs producer acp1 admin slot load --slot 1 --card axon/synapse/RRS18-1601/acp1`.

## Stacked on

PR #281 (slot state machine). Merge order: Phase 0 (#267-#272) → #273 → #274 → #275 → #276 → #277 → #278 → #279 → #280 → #281 → this.

## Issue refs

Advances #260. Closes on the eventual PR-to-main when the ACP1 epic bundle lands together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #281, opened 2026-05-05
feat(acp1): provider slot state machine + realistic timings (advances #259)

by-rune:
## Summary

- New `slotStateMachine` on the ACP1 provider tracks per-slot cancellable cascades.
- `CascadeInsert` drives `no_card -> powerup -> boot -> present`, emitting a frame-status announce at each phase.
- `CascadeExtract` drives `present -> removed -> no_card` immediately.
- Two timing modes: `InsertTimingReal` (1500ms powerup + 2500ms boot, spec-typical) and `InsertTimingFast` (50ms per phase, CI-friendly).
- New admin verbs: `slot.insert --slot N` (kicks off the cascade) and `slot.extract --slot N` (immediate).
- Producer CLI: `--insert-timing real|fast` (default `real`) plumbs to `SetInsertTiming` on the server.
- 9 unit tests covering parse, durations, full-cascade reaches present, per-phase observation, cancellation, immediate extract, pre-emption, timing mutation.

## State machine (per spec p.20)

```text
no_card  -> powerup : immediate (admin slot.insert)
powerup  -> boot    : after powerupDuration (real=1500ms / fast=50ms)
boot     -> present : after bootDuration   (real=2500ms / fast=50ms)
present  -> removed : immediate (admin slot.extract)
removed  -> no_card : immediate (admin slot.extract follow-up)
*        -> error   : immediate (admin slot.state --state error, lands in #258)
```

Cancellation:
- Server stop cancels parent context — pending timers drain.
- New `slot.insert` for the same slot pre-empts the in-flight cascade.
- `slot.extract` mid-cascade pre-empts the insert cascade.

## Why two timings

Real timings make the emulator visually match a Synapse hot-plug for live testing. Fast mode cuts the cascade to ~100ms total so the `feat/acp1-full` integration suite can drive `slot.insert` repeatedly without burning seconds on every test.

## Test plan

- [x] `go test ./internal/acp1/provider/... -count=1 -run "TestCascade|TestParseInsertTiming|TestInsertTiming|TestSlotStateMachine"` — 9 cases green.
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Live test: drive `admin slot.insert` over the wire and watch the cascade in `dhs consumer acp1 watch`.

## Stacked on

PR #280 (admin CLI). Merge order: Phase 0 (#267-#272) → #273 → #274 → #275 → #276 → #277 → #278 → #279 → #280 → this.

## Issue refs

[...]

PR #285, opened 2026-05-05
feat(acp1): provider VIP-aware broadcast source (advances #263)

by-rune:
## Summary

- Pin the broadcast socket's `LocalAddr` to the listener's bind IP when `--bind <ip>` is set. Multi-instance emulators on different VIPs now appear as distinct `From:` addresses to consumers.
- New `--bind` alias for `--host` on `dhs producer acp1 serve`. Both spellings work; `--bind` is the canonical name from the design discussion.
- Default `--bind 0.0.0.0` leaves the broadcast `LocalAddr` nil so legacy behaviour (kernel picks source) is preserved.
- 2 unit tests pin the LocalAddr behaviour on loopback.

## Why this matters

```
host A: --bind 10.6.239.200 --tree rack-A.json
host A: --bind 10.6.239.201 --tree rack-B.json
```

Two emulators, same machine, two VIPs. Without source-IP pinning, both broadcasts source from whichever IP the kernel default-routes — consumer sees `From: 10.6.239.X` for both, can't tell them apart. With this PR, each broadcast carries the bind IP, and `dhs consumer acp1 discover` lists both as distinct origins.

## Multi-VIP integration

The real multi-VIP scenario needs multiple non-loopback IPs configured on different ifaces. The LXC rig already provides this; CI cannot. Unit tests assert the bind logic on loopback (which is the only thing we can portably test); integration tests on the LXC rig verify the cross-iface case.

## Test plan

- [x] `go test ./internal/acp1/provider/... -count=1 -run TestVIPBind` — 2 cases green.
- [x] `go build ./...` clean.
- [x] golangci-lint clean.
- [ ] Live: two emulator instances on `10.6.239.200` and `.201`, `dhs consumer acp1 discover` lists both.

## Stacked on

PR #284 (fuzz verb). Merge order: Phase 0 (#267-#272) → #273 → #274 → #275 → #276 → #277 → #278 → #279 → #280 → #281 → #282 → #283 → #284 → this.

## Issue refs

Advances #263. Closes on the eventual PR-to-main when the ACP1 epic bundle lands together.

### DIFF
--- a/cmd/dhs/cmd_producer.go
+++ b/cmd/dhs/cmd_producer.go
@@ -41,7 +41,8 @@ func runProducer(ctx context.Context, protoName string, args []string) error {
 	var (
 		treePath      = fs.String("tree", "", "path to canonical tree.json (required)")
 		port          = fs.Int("port", 0, "TCP listen port (0 = plugin default)")
-		host          = fs.String("host", "0.0.0.0", "TCP listen host")
+		host          = fs.String("host", "0.0.0.0", "TCP/UDP listen host (alias: --bind)")
+		bind          = fs.String("bind", "", "alternate spelling of --host. e.g. --bind 10.6.239.200 binds the listener AND pins the broadcast source IP to the VIP, so multi-instance emulators on the same machine appear as distinct From: addresses to consumers (#263).")
 		logLevel      = fs.String("log-level", "info", "log level: debug, info, warn, error")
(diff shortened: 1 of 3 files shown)


by-rune:
Sub-tracker complete — work landed in main as `16f7335`. Closing the tracker; any residual scope continues on the parent issue if still open.

PR #328, opened 2026-05-08
fix(acp2-provider): always emit pid=6 (string_max_length) on String objects

### LINKED ISSUE #309: fix(acp2-provider): always emit pid=6 (string_max_length) on String objects
**Spec reference:** `internal/acp2/assets/acp2_protocol.docx` §"Property fields" matrix. Pid 6 (string_max_length) is required (Y) on the String row. §5.4 row 6: `data=0, plen=6, body=u16 len + u16 pad`.

**Current behavior:** `internal/acp2/provider/encoder.go:144` only emits pid=6 when `maxLenHint(e.param) > 0`. When the canonical Parameter omits the max-length hint, our wire skips pid=6 entirely. Spec-strict consumers (and real Neuron) always carry it.

**Required behavior:** emit pid=6 on every String reply. When canonical max-length is unset, fall back to a sensible default (256 per spec §"Requirements" — "A string type should support strings up to 256 bytes, configurable in menu") OR pull from the actual `len(value)` + buffer.

**Acceptance criteria:**
- Byte-table unit test asserts pid=6 present on every String reply, regardless of canonical hint.
- Wire diff vs real Neuron capture on a String object whose canonical fixture lacks an explicit max-length = 0 bytes.
- Cerebrum + VSM round-trip green.

Closes part of #301.

by-rune:
## Summary
Provider's `buildProperties` now always emits **pid=6 (string_max_length)** on every String reply per ACP2 spec §"Property fields" matrix (Y on String row). Previously skipped when `maxLenHint(canonical) == 0`.

When canonical lacks an explicit hint, falls back to **256 bytes** per spec §"Requirements" line 173:
> A string type should support strings up to a 256 bytes, configurable in menu.

Wire shape (§5.4 row 6): `data=0, plen=6, body=u16 BE len + u16 pad`.

## Test plan
- [x] `TestBuildProperties_String_NoCanonicalMaxLen` — asserts pid=6 present with default value 256 when canonical hint absent
- [x] `TestBuildProperties_String_WithMaxLen` (existing) — continues to pass with canonical-supplied value (16)
- [x] All ACP2 tests green
- [x] Build clean
- [ ] CI green
- [ ] Cerebrum + VSM Studio re-walk: every String value (e.g. `IDENTITY > User Label 1`, `MANAGEMENT PORT > DNS Domain`, etc.) carries pid=6 — no parser warnings.

## Risk / blast radius
Wire frame for every String reply gains a single 8-byte property record (header 4 + body 2 + pad 2). Consumer walker already handles pid 6 (`PIDStringMaxLength` case in `walker.go:185`).

Closes #309. Stacks under `feat/acp2-strict-spec-closeout`.

### DIFF
--- a/internal/acp2/provider/encoder.go
+++ b/internal/acp2/provider/encoder.go
@@ -10,6 +10,13 @@ import (
 	"dhs/internal/acp2/codec"
 )
 
+// stringMaxLengthDefault is the spec-stated upper bound for ACP2
+// string objects (§"Requirements" line 173: "A string type should
+// support strings up to 256 bytes, configurable in menu"). Used as
+// the pid=6 (string_max_length) value when the canonical fixture
+// omits an explicit hint.
+const stringMaxLengthDefault = 256
+
 // buildProperties assembles the ACP2 property list a get_object reply
 // must carry for one entry. Per spec §"Property IDs" the ordering is
 // not strictly required but consumers generally expect:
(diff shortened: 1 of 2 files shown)


by-rune:
Applied via cherry-pick on `feat/acp2-strict-spec-closeout`: 18856d0. Subsumed by #350 (bundled §5.1 pid emission, including pid=6 always on String). Cherry-picked onto closeout. Issue #309 will be closed by the umbrella PR.

PR #342, opened 2026-05-08
feat(acp2-provider): sanitize labels to spec charset per spec §Versions

### LINKED ISSUE #318: feat(acp2-provider): validate label charset [A-Za-z0-9 -] per spec section Versions
**Spec reference:** `internal/acp2/assets/acp2_protocol.docx` §"Versions" line 357-371:
> Object labels only may contain out of a-z, A-Z, 0-9, ' ' and '-'.

This is a wire-level constraint: any pid=2 (label) string emitted on the wire MUST contain only those characters.

**Current behavior:** `internal/acp2/provider/tree.go::flatten` accepts any UTF-8 label from `tree-fresh.json` and passes it through to `propStringData0(codec.PIDLabel, e.label)`. Real-Neuron-derived labels generally comply, but engineered/fixture trees can carry forbidden characters (`.`, `:`, `/`, `_`, accented chars, etc.) without warning.

**Required behavior:** at tree load time, validate every label against `^[A-Za-z0-9 \-]*$`. On violation:
- Reject the tree with a clear error pointing at the bad obj-id and label, OR
- Emit a `compliance.Event` and sanitize (replace forbidden chars with `-`) — preferred.

**Acceptance criteria:**
- Loader test: tree with a label "User_Label" or "DNS.Primary" raises a compliance event.
- Tree with all-compliant labels loads silently.
- Wire-level invariant test: every pid=2 byte sequence in a serving session passes the regex.

Closes part of #301.

by-rune:
## Summary
Provider's `flatten` now runs every label through `sanitizeLabel` before storing it on the in-memory entry. ACP2 spec §"Versions" line 357 restricts labels to `[A-Za-z0-9 \-]`; any other character is replaced with `-` so our wire stays spec-clean.

## Why this matters
Real EVS Neuron firmware emits non-compliant labels like `"ROOT_NODE_V2"` (underscore) and `"1080p23.98"` (dot). Per `feedback_acp2_spec_pdfs_only` the docx is the only authority, and per `feedback_no_workaround` we match it strictly — even when the source device deviates.

## Behaviour
| Input | Output | Why |
|---|---|---|
| `ROOT_NODE_V2` | `ROOT-NODE-V2` | underscore not in spec set |
| `Sample.Rate` | `Sample-Rate` | dot not in spec set |
| `MAC:01:23` | `MAC-01-23` | colon not in spec set |
| `Café` | `Caf-` | non-ASCII stripped |
| `""` / `___` | `obj` | empty / all-invalid → placeholder |
| `User Label 1` | identity | already compliant |
| `PSU - NPU0500-B` | identity | dash + space allowed |
| `CHANNEL 01` | identity | already compliant |

The unmodified label stays on the canonical element. Only the in-memory `entry.label` (which feeds pid=2 emission) is sanitised — other tooling reading the canonical export still sees the original.

## Test plan
- [x] `TestSanitizeLabel` — 14 sub-cases (compliant identity, common Neuron deviations, non-ASCII, edge cases).
- [x] `TestIsSpecLabelByte` — explicit allow + disallow rune lists.
- [x] All ACP2 + repo tests green (`go test ./...` zero failures).
- [x] Build clean
- [ ] CI green
- [ ] Integration: Cerebrum / VSM render tree against the renamed labels (e.g. `ROOT-NODE-V2`). May need downstream-config updates if anything string-matched the old name.

## Notes for reviewer
This is a **wire-visible** change — the strings Cerebrum / VSM see for non-compliant source labels will differ. Strict-spec is the right call per binding memory rules; if the real-device-compat goal outweighs strict spec, we relax this.

Closes #318. Stacks under `feat/acp2-strict-spec-closeout`.

### DIFF
--- a/internal/acp2/provider/label_charset_test.go
+++ b/internal/acp2/provider/label_charset_test.go
@@ -1,0 +1,71 @@
+package acp2
+
+import "testing"
+
+// TestSanitizeLabel pins the spec invariant from acp2_protocol.docx
+// §"Versions" line 357: object labels MUST contain only characters
+// from `[A-Za-z0-9 \-]`. Any other character is replaced with `-`
+// before the label hits the wire (pid=2). The unmodified label stays
+// on the canonical element so other tooling can see the original.
+func TestSanitizeLabel(t *testing.T) {
+	cases := []struct {
+		name  string
+		in    string
+		out   string
+		dirty bool // whether the function should have changed the input
+	}{
+		// Compliant labels round-trip identity.
+		{"all-lower", "abcdef", "abcdef", false},
+		{"mixed-case-digits", "Card1234", "Card1234", false},
+		{"with-space", "User Label 1", "User Label 1", false},
+		{"with-dash", "Slot-A", "Slot-A", false},
+		{"compound", "PSU - NPU0500-B", "PSU - NPU0500-B", false},
+		{"channel-N", "CHANNEL 01", "CHANNEL 01", false},
+
+		// Real-Neuron deviations — spec violations sanitised.
+		{"underscore", "ROOT_NODE_V2", "ROOT-NODE-V2", true},
+		{"dot", "Sample.Rate", "Sample-Rate", true},
+		{"colon", "MAC:01:23", "MAC-01-23", true},
+		{"slash", "A/B", "A-B", true},
+		{"equal-sign", "Mode=Auto", "Mode-Auto", true},
+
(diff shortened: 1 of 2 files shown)


by-rune:
Applied via cherry-pick on `feat/acp2-strict-spec-closeout`: cd7af27. Label charset sanitize applied via cherry-pick on feat/acp2-strict-spec-closeout. Issue #318 will be closed by the umbrella PR.

PR #369, opened 2026-05-09
feat(acp2,cli): auto-reconnect on session loss with exponential backoff (closes #367)

### LINKED ISSUE #367: feat(consumer): auto-reconnect on session loss (cross-protocol, plugin-internal)
## Status

Today's investigation of #365 (acp2 keep-alive) confirmed the keep-alive watchdog correctly detects silence (TCP-up but device frozen / SIGSTOP scenario). However it does **not** cover the case where the TCP socket is fully closed: producer crash, SIGKILL, network drop, controller restart.

In that case:

- read loop sees EOF / `connection closed`, calls `closeErr`, returns;
- session.done channel closes;
- plugin still holds a non-nil `p.session` pointer to a defunct Session;
- watch verb's main loop sits blocked on `select { case <-ctx.Done(); case ev := <-events }` with no events arriving;
- operator sees a frozen table — no values, no announces, no recovery.

VSM Studio + Cerebrum both reconnect automatically when the device comes back. Our watch verb should match.

## Required behaviour (cross-protocol)

When the session goes from connected → disconnected (read loop EOF / write fail / dead-man past N×timeout):

1. Mark all in-memory tree values' freshness as **`stale`** (was live in this session, now uncertain). Schema retained.
2. Fire `compliance.SessionLost` event.
3. Start a reconnect goroutine with **exponential backoff**: 1 s, 2 s, 4 s, 8 s, 16 s, capped at 30 s. Configurable via `--reconnect-backoff-cap`.
4. On each retry, attempt `Connect(host, port)` (full handshake re-runs).
5. On success:
   - re-issue `EnableProtocolEvents([proto])` (ACP2 specifically) so announces resume;
   - re-issue any active `Subscribe` registrations the operator had;
   - flip every cached value's freshness from `stale` back to → wait for first announce / probe → `live`;
   - log `info` "session reconnected after N attempts (M elapsed)".
6. Until success: continue retrying. Operator can `--max-reconnect-attempts N` (default `0` = unlimited).

## Key invariants

- **Schema retained across reconnect.** No re-walk required for same-identity card. Identity-shift triggers re-walk per existing #298 contract.
- **`watch` does not exit.** It blocks waiting for events; reconnect happens transparently behind it. Operator only loses traffic during the gap; the table re-fills as announces arrive.
- **No spam during outage.** First disconnect logs `warn`; subsequent reconnect attempts log at `debug`. Successful reconnect logs `info` with summary.
- **Per-protocol per-plugin** — each connector handles its own reconnect (no shared dispatcher logic). Mirrors the keep-alive contract pattern.

## Concrete changes (plugin-internal each, one PR per connector)

ACP2 first since the gap is freshest:

| File | Change |
|---|---|
| `internal/acp2/consumer/reconnect.go` (new) | `reconnectLoop(ctx, backoffCap, maxAttempts)` goroutine; `Plugin.onSessionLost` callback hook fired by the read loop's EOF path; freshness=stale flip via the existing tree cache. |
| `internal/acp2/consumer/session.go` | `Session.OnLost(func(err error))` registration; readLoop calls it on close. |
| `internal/acp2/consumer/plugin.go` | `Connect` registers `onSessionLost`; reconnect goroutine started lazily on first loss; subscribe registry re-played after handshake re-runs. |
| `internal/acp2/consumer/reconnect_test.go` (new) | Unit: simulate EOF → assert reconnect goroutine starts, exponential backoff respected, schema retained, freshness=stale set then live again. |

Then ACP1, Ember+, Probel, OSC, TSL — each gets its own PR mirroring the same pattern.

## Cross-protocol contract

Add `protocol.Reconnecter` interface (optional like `KeepAliver`):

```go
// Reconnecter is the optional capability a plugin exposes when it
// supports automatic session re-establishment after loss. Plugins that
// don't implement it are unaffected; the harness CLI just doesn't
// surface the --reconnect-* flags for them.
type Reconnecter interface {
    SetReconnect(cfg ReconnectConfig)
}

type ReconnectConfig struct {
    Initial      time.Duration // first backoff (default 1s)
    Cap          time.Duration // max backoff (default 30s)
    MaxAttempts  int           // 0 = unlimited
    Disabled     bool          // skip reconnect entirely (current behaviour)
}
```

CLI flags propagate identically across protocols (cmd_watch.go, cmd_health.go).

## Out of scope

- Replaying events that fired during the outage. Standard Prometheus / Loki / etc.: gaps in time series are expected and acceptable. Don't try to back-fill.
- Identity-shift detection during reconnect — already handled by #298 contract (different `CardIdentity` → re-walk).
- Hot-plug `auto-walk-on-plug` default flip — separate concern.

## Acceptance

- [ ] Per-connector reconnect implementation merged (one PR per protocol).
- [ ] Live LXC test: SIGKILL the producer, watch the consumer reconnect after producer restart, verify table re-fills with `live` freshness.
- [ ] Live LXC test: pull the network cable equivalent (drop the producer's host route via `ip route del`), verify retry backoff, verify recovery on route restore.
- [ ] `audits/connector-readiness.md` row updated for each connector.

## Reference

Today's #365 verification 2026-05-10 caught the SIGSTOP case but explicitly didn't cover SIGKILL / TCP RST. tshark capture earlier confirmed Cerebrum's idle traffic = `AN2 GetSlotInfo` every 5 s; we don't have wire evidence of Cerebrum's reconnect cadence yet — capturing that on the next pass is part of the ACP2 implementation PR.

by-rune:
Stacks on #366 (acp2 keep-alive). Live-verified end-to-end:

```
T+0   watch connects
T+13  pkill -9 producer  -> session lost
T+14  attempt #1 delay=1s
T+18  attempt #2 delay=2s
T+24  attempt #3 delay=4s
T+34  attempt #4 delay=8s   <- producer restarted via setsid
T+52  attempt #5 delay=16s
T+52  acp2: connected + reconnect: succeeded attempt=5
```

Exponential backoff (1s -> 2s -> 4s -> 8s -> 16s, cap 30s), subscriptions replayed, watch resumed without operator intervention.

Closes #367.

### DIFF
--- a/cmd/dhs/common.go
+++ b/cmd/dhs/common.go
@@ -37,6 +37,9 @@ type commonFlags struct {
 	timeout           time.Duration
 	keepalive         time.Duration
 	keepaliveTimeout  time.Duration
+	reconnect         bool
+	reconnectCap      time.Duration
+	reconnectAttempts int
 	verbose           bool
 	logLevel          string
 	capture           string
@@ -77,6 +80,16 @@ func addCommonFlags(fs *flag.FlagSet) *commonFlags {
 		"keep-alive dead-man threshold (0 = 3× --keepalive; -1 = never "+
 			"declare session dead). Watch verb shows freshness=cache once "+
 			"this elapses without rx; values stay decoded against the cached schema.")
+	fs.BoolVar(&cf.reconnect, "reconnect", true,
+		"on session loss (TCP close / device crash / network drop), retry "+
+			"the connection with exponential backoff. Plugins that don't "+
+			"implement protocol.Reconnecter ignore this flag.")
+	fs.DurationVar(&cf.reconnectCap, "reconnect-cap", 0,
+		"cap on the exponential backoff between reconnect attempts "+
+			"(0 = plugin default 30s)")
+	fs.IntVar(&cf.reconnectAttempts, "reconnect-max-attempts", 0,
+		"maximum reconnect attempts before giving up "+
+			"(0 = unlimited; the watch verb keeps trying)")
 	fs.BoolVar(&cf.verbose, "verbose", false, "debug log output (shortcut for --log-level debug)")
 	fs.StringVar(&cf.logLevel, "log-level", "info", "log level: trace, debug, info, warn, error, critical")
 	fs.StringVar(&cf.capture, "capture", "",
@@ -214,6 +227,16 @@ func connect(ctx context.Context, host string, cf *commonFlags) (protocol.Protoc
 		ka.SetKeepAlive(protocol.KeepAliveConfig{
 			Interval: cf.keepalive,
 			Timeout:  cf.keepaliveTimeout,
+		})
+	}
+
+	// Reconnect — same plugin-optional pattern as KeepAliver. Plugins
+	// that don't implement Reconnecter silently ignore.
+	if rc, ok := plug.(protocol.Reconnecter); ok {
+		rc.SetReconnect(protocol.ReconnectConfig{
+			Cap:         cf.reconnectCap,
+			MaxAttempts: cf.reconnectAttempts,
+			Disabled:    !cf.reconnect,
 		})
 	}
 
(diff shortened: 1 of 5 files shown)


by-rune:
Reverting — Subscribe replay logs as 'replayed' but post-reconnect the wrapper doesn't deliver announces to the watch loop. Approach is unsound; will redesign in a follow-up that doesn't risk breaking the working keepalive path.

Issue #415, opened 2026-05-12
fix(cli/import): skip walk for ACP2/EmberPlus apply (CSV carries obj-id)


by-rune:
## Problem

`dhs consumer acp2 import <host> --file X.csv` (without `--dry-run`) still walks the full slot before iterating CSV rows. On ACP2 slot 1 (~49 827 objects) that walk is the dominant cost: a 704-row CSV apply takes ~5 minutes on the .103 producer even though the 704 actual SetValues are sub-second.

#413 fixed the dry-run path. This issue fixes the apply path for ACP2 (and EmberPlus).

## Why the walk can go

For ACP2, the CSV `id` column carries the wire obj-id; importer.go:146 sets `req.ID = obj.ID`. ACP2 `SetValue` ([internal/acp2/consumer/plugin.go:524](internal/acp2/consumer/plugin.go#L524)) already falls back to `fetchObjectMeta` (single `get_object`) when no walked tree is loaded. So a walk-less apply does 1 extra get_object per row (~ms each) — 704 rows = ~5 s, not 5 min.

Same applies to EmberPlus (path-keyed, no walk needed).

ACP1 must keep the walk — its `SetValue` requires the walked tree for typed encoding (no per-object fallback).

## Fix

In `internal/export/importer.go` `Apply`, gate the walk by `s.Device.Protocol`:

```go
needsWalkForApply := !dryRun && (s.Device.Protocol == acp1)
if needsWalkForApply {
    if _, err := plug.Walk(ctx, dump.Slot); err != nil { ... }
}
```

## Done-when

- [ ] Walk in Apply gated to ACP1 only (acp2/emberplus skip)
- [ ] Unit tests:
  - `TestApply_ACP2_ApplySkipsWalk` — counting mock, snapshot device.protocol=`acp2`, dryRun=false → walkCalls==0, setCalls==N
  - `TestApply_ACP1_ApplyStillWalks` — protocol=`acp1`, dryRun=false → walkCalls==1
- [ ] Live smoke on .103 producer: 704-row `neuron-senders.csv` apply finishes in < 10 s (was ~5 min)
- [ ] All existing import tests still green
- [ ] Sign-off from @yboujraf

## Out of scope

- ACP1 walk-less apply (needs per-object fallback in ACP1 SetValue) — separate issue
- Validate / strict / new skip reasons — #410
- `defer recover()` — #410

by-rune:
[...]

PR #416, opened 2026-05-12
fix(cli/import): skip walk on apply for ACP2/EmberPlus (#415)

by-rune:
## Summary

Follow-up to #414. Apply was unconditionally walking the slot before iterating snapshot rows. For ACP2/EmberPlus that walk is dead weight: the CSV carries `id` (acp2) or `oid/path` (emberplus), and `SetValue` has a per-object meta-fetch fallback. ACP1 still needs the walk (its `SetValue` requires the walked tree for typed encoding).

Gate: walk only when `!dryRun && Protocol == acp1`.

## Measured (704-row `neuron-senders.csv` against ACP2 producer on .103)

| Path | Before | After |
|---|---:|---:|
| `--dry-run` (from #414) | minutes | 26 ms |
| Real apply (this PR) | **~5 min** | **360 ms** |

## Test plan

- [x] `TestApply_DryRunSkipsWalk`           dryRun=true, acp2  -> walks=0 sets=0
- [x] `TestApply_ACP2_ApplySkipsWalk`       dryRun=false, acp2 -> walks=0 sets=2
- [x] `TestApply_EmberPlus_ApplySkipsWalk`  dryRun=false, ep   -> walks=0 sets=2
- [x] `TestApply_ACP1_ApplyWalks`           dryRun=false, acp1 -> walks=1 sets=2 (no regression for ACP1)
- [x] `go test ./internal/export/...` green
- [x] `go build ./...` green; `golangci-lint` green (pre-commit)
- [x] Live smoke on .103 producer: `time dhs consumer acp2 import 127.0.0.1 --slot 1 --file /tmp/neuron-senders.csv` -> `real 0m0.360s`, output `applied 704, skipped 0, failed 0`

## Dependencies

Branches on top of [feat/cli-import-no-walk-413](https://github.com/by-openclaw/go-acp/pull/414). Merge #414 first, then this rebases cleanly.

Refs #415. (No auto-close; close requires explicit approval.)

### DIFF
--- a/internal/export/importer.go
+++ b/internal/export/importer.go
@@ -88,14 +88,17 @@ func LoadSnapshot(path string) (*Snapshot, error) {
 func Apply(ctx context.Context, plug protocol.Protocol, s *Snapshot, dryRun bool) (*ImportReport, error) {
 	rep := &ImportReport{DryRun: dryRun}
 
+	// Walk is only needed when the protocol's SetValue requires a
+	// pre-populated label/type map. ACP1 does (SetValue errors out
+	// without a tree — see internal/acp1/consumer/plugin.go:597).
+	// ACP2 + EmberPlus have per-object meta-fetch fallbacks, so the
+	// walk is dead weight when the CSV already carries obj-id (acp2)
+	// or OID/path (emberplus). Dry-run never reaches SetValue at all,
+	// so the walk is dead weight there for every protocol.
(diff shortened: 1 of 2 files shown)


yboujraf:
you don't need for acp1 to walk through, acp1 uses as well objectID !!!

yboujraf (approved):

PR #446, opened 2026-05-16
fix(emberplus/consumer): set --value silently coerces unparseable input to zero (closes #445)

### LINKED ISSUE #445: fix(emberplus/consumer): set --value silently coerces unparseable input to zero
## Bug

\`coerceStringToTyped\` in \`internal/emberplus/consumer/plugin.go:2295\` swallows \`strconv.ParseInt\` / \`ParseFloat\` / \`ParseUint\` errors. On parse failure, the typed field (\`val.Int\` / \`val.Float\` / \`val.Uint\`) stays at its Go zero value, the wire encodes that zero, the provider stores the zero, and the consumer reports \`confirmed value = 0\` as if the SET succeeded.

Result: any unparseable string against a typed Parameter is silently coerced to 0 (or false / 0.0). Strict-spec violation per \`feedback_no_workaround\` — the consumer must reject invalid input, not silently rewrite it.

## Reproduction

Against the integration demo (port 9100, identity \`dhs-emberplus-integration@1.0.0\`), \`nToN.targetParams.0.gain\` is an Integer Parameter (range -1000..0, step 1):

\`\`\`powershell
.\bin\dhs.exe consumer emberplus set 127.0.0.1 --port 9100 \`
    --dm dhs-emberplus-integration@1.0.0 \`
    --path dhs-emberplus-integration.nToN.targetParams.0.gain \`
    --value -25.5
# confirmed value = 0       ← BUG: should error on non-integer input
\`\`\`

For comparison, \`--value -2\` (valid integer) works correctly: \`confirmed value = -2\`.

## Scope

\`internal/emberplus/consumer/plugin.go\`:

- \`coerceStringToTyped\` returns \`error\` instead of \`void\`. Each \`Parse*\` branch returns a wrapped error on failure with the rejected string + the expected kind.
- The call site at line 727 (inside SetValue) propagates the error as a \`protocol.ValidationError\` (exit code 2 — same shape as cmd_set already uses for usage errors).

## Acceptance

- Unit test: \`set --value -25.5\` against an Integer Parameter returns a non-nil error containing the rejected string \`"-25.5"\`.
- Live: same set against integration demo returns \`error: invalid value "-25.5" for integer parameter\` and exit code 2; the parameter's value is **unchanged** (still 0, not silently overwritten with 0).
- No regression: every existing roundtrip in \`scripts/emberplus/verify-emberplus-integration.ps1\` still 22/22 PASS.

## Out of scope

- ACP1 / ACP2 codecs handle parse-failure inside their wire encoders separately. Audit there is a follow-up if the same anti-pattern exists.
- Whether the demo's \`gain\` Parameter *should* be Real (float) instead of Integer for real-world feel — separate UX discussion, not this bug.

## Refs

- Discovered during cross-client validation of #441 + #437 (post-merge testing on EmberPlusView + Cerebrum)
- \`feedback_no_workaround\` — spec-strict, no silent coercion
- \`internal/emberplus/CLAUDE.md\` §Spec-strict, no-workaround posture

by-rune:
## Summary

\`coerceStringToTyped\` in \`internal/emberplus/consumer/plugin.go\` was swallowing \`strconv.ParseInt\` / \`ParseFloat\` / \`ParseUint\` errors: on bad input the typed field (\`val.Int\` / \`val.Float\` / \`val.Uint\` / \`val.Enum\`) stayed at its Go zero value, the wire encoded that zero, the provider stored it, and the consumer reported \`confirmed value = 0\` as if the SET had succeeded.

Caught during cross-client validation post-#441 merge — \`set --value -25.5\` against an Integer Parameter (range -1000..0, step 1) silently rewrote the value to 0 instead of erroring.

## Fix

- \`coerceStringToTyped\` returns \`*protocol.ValidationError\` on parse failure for KindInt / KindUint / KindFloat / KindEnum. Reason carries the rejected string for operator debuggability.
- Call site at \`plugin.go:727\` (inside SetValue) propagates; CLI exits with status 2 (the existing usage-error convention via \`exitCode()\` in \`cmd/dhs/main.go\`).
- KindBool stays permissive — operators expect \`"true"\` / \`"1"\` / \`"yes"\` → true and anything else → false. The doc-comment now spells that out so future tightening doesn't accidentally regress operator-friendly bool input.
- String passthrough and empty-string guard unchanged.

## Live repro on integration demo

\`\`\`
$ set ... --value -25.5
error: validation: value: invalid integer \"-25.5\"  (exit 2)

$ set ... --value abc
error: validation: value: invalid integer \"abc\"    (exit 2)

$ set ... --value -2
confirmed value = -2                                  (exit 0)
\`\`\`

## Test plan

- [x] New \`coerce_test.go\` pins 23 cases:
  - 9 reject paths (KindInt / KindUint / KindFloat / KindEnum × their main invalid forms — letters, decimals on int, negative on uint, overflow on enum, etc.)
  - 10 accept paths (happy cases per kind)
  - 3 KindBool permissive cases
  - 1 KindString + empty passthrough case
- [x] \`go test ./internal/emberplus/...\` — all green
- [x] \`go build ./...\` clean
- [x] Live: rebuilt \`bin/dhs.exe\` from this branch, cycled producer, reproduced the failing case → now errors with exit code 2; sanity check on valid input \`-2\` unchanged
- [x] \`scripts/emberplus/verify-emberplus-integration.ps1\` 22/22 PASS on the integration demo (transient script-side race observed once, passes deterministically on retry — same flake we've seen since the verifier landed in #441)

## Out of scope

- ACP1 / ACP2 codecs handle parse-failure inside their own wire encoders; audit there is a follow-up if the same anti-pattern exists (separate issue).
- Whether the demo's \`gain\` Parameter *should* be Real (float) instead of Integer for real-world feel — separate UX discussion.

Closes #445

### DIFF
--- a/internal/emberplus/consumer/coerce_test.go
+++ b/internal/emberplus/consumer/coerce_test.go
@@ -1,0 +1,165 @@
+package emberplus
+
+import (
+	"errors"
+	"testing"
+
+	"dhs/internal/protocol"
+)
+
+// TestCoerceStringToTyped_RejectsInvalidInput pins #445: unparseable
+// --value strings must surface as protocol.ValidationError, not be
+// silently coerced to the Go zero value of the typed field. Each row
+// asserts that the parse failure (a) returns a non-nil error, (b)
+// types as *protocol.ValidationError so exitCode() yields 2, and (c)
+// the rejected string appears in Reason for operator debuggability.
+func TestCoerceStringToTyped_RejectsInvalidInput(t *testing.T) {
+	cases := []struct {
+		name string
+		kind protocol.ValueKind
+		str  string
+	}{
+		{"int with decimal", protocol.KindInt, "-25.5"},
+		{"int with letters", protocol.KindInt, "abc"},
+		{"int empty-after-sign", protocol.KindInt, "-"},
+		{"uint with negative", protocol.KindUint, "-1"},
+		{"uint with letters", protocol.KindUint, "xyz"},
+		{"float with letters", protocol.KindFloat, "3.14abc"},
+		{"float garbage", protocol.KindFloat, "not-a-float"},
+		{"enum with letters", protocol.KindEnum, "Low"},
+		{"enum overflow uint8", protocol.KindEnum, "999"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			val := &protocol.Value{Kind: tc.kind, Str: tc.str}
+			err := coerceStringToTyped(val)
+			if err == nil {
+				t.Fatalf("kind=%v str=%q: expected error, got nil (val=%+v)", tc.kind, tc.str, val)
+			}
+			var verr *protocol.ValidationError
+			if !errors.As(err, &verr) {
+				t.Errorf("kind=%v str=%q: expected *protocol.ValidationError, got %T (%v)", tc.kind, tc.str, err, err)
+			}
+			if verr != nil && !contains(verr.Reason, tc.str) {
+				t.Errorf("kind=%v: Reason %q should mention rejected input %q", tc.kind, verr.Reason, tc.str)
+			}
+			// Typed field MUST stay at zero — no silent coerce.
+			if val.Int != 0 || val.Uint != 0 || val.Float != 0 || val.Enum != 0 {
+				t.Errorf("kind=%v str=%q: typed fields should be zero on error, got %+v", tc.kind, tc.str, val)
+			}
+		})
+	}
+}
+
+// TestCoerceStringToTyped_AcceptsValidInput pins the happy paths so a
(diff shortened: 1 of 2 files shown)


by-rune:
Closing in favour of #453.

PR #446 was authored against the pre-rename `internal/protocol/` package. PR #452 (merged in v0.12.0) renamed `internal/protocol/` â†’ `internal/consumer/` for symmetric identity with `internal/provider/`. Rather than rebase this PR's diff through the import churn, #453 re-applies the same `coerceStringToTyped` fix on top of current main using the post-rename names (`consumer.ValidationError` / `consumer.Value` / `consumer.KindInt` etc.).

Same 23-case test pinning, same SetValue propagation path, same operator-friendly KindBool semantics. Refs #445 stays open until merge.

PR #618, opened 2026-06-22
test(ansible): cross-OS dhs_verbtest role + probel-sw08p all-verbs template + site.yml

by-rune:
Foundation for cross-OS (Windows + Linux + macOS, ADR-0016) Ansible verb testing of the REAL dhs binary (dhs.exe on Windows). Answers: yes, tests drive the actual binary, on all OSes. Role ansible/roles/dhs_verbtest: builds the binary (go build -> dhs/dhs.exe), starts the loopback producer fire-and-forget (async/poll:0 + bounded TTL; the async reaper auto-expires it -> NO per-OS nohup/pkill/Stop-Process, zero persistent change/idempotent), wait_for/win_wait_for for race-safety, then runs+asserts a connector's FULL verb list. All OS variance (command vs win_command, wait_for vs win_wait_for, .exe ext, temp dir) is confined to the role via ansible_os_family dispatch; connector plays carry zero per-OS code (supply only vt_proto/vt_serve_argv/vt_verbs). probel-sw08p-verbs.yml wires all 23 verbs (each cited to cmd/dhs source) as the template. site.yml orchestrates unit-tests.yml (tag unit) + verb plays (tags smoke/<proto>), OS-targeted via -e vt_target; old site.yml kept as site-legacy.yml; DHS_BIN overrides the binary path. Authored from source + the repo's roles/dhs_acp1 per-OS pattern; go build clean, YAML parses; NOT executed here (no ansible/WinRM on the dev host) — the Windows win_command-async path needs live confirmation. The other 7 connectors replicate via the role as a data-only follow-up. Co-Authored-By Claude Opus 4.8.

### DIFF
--- a/ansible/playbooks/probel-sw08p-verbs.yml
+++ b/ansible/playbooks/probel-sw08p-verbs.yml
@@ -1,0 +1,464 @@
+---
+# probel-sw08p-verbs.yml — the CROSS-OS verb-matrix smoke play for Probel
+# SW-P-08 / SW-P-88. This is the TEMPLATE the other 7 connectors copy: it
+# carries ONLY data (the producer serve args + the full consumer verb list)
+# and delegates ALL execution + every per-OS decision to the dhs_verbtest
+# role. There is no `command` vs `win_command` here, no wait_for vs
+# win_wait_for, no .exe handling — the role owns all of that (ADR-0016).
+#
+# The role:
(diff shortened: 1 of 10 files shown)


by-rune:
Superseded by the deploy-over-WinRM version: the dhs_verbtest role here wrongly built the binary on the target, but win11 has no Go toolchain. The replacement cross-builds on the control node and deploys via copy/win_copy, matching the proven dhs_acp1 WinRM pattern.

PR #620, opened 2026-06-22
test(ansible): cross-OS verb plays for all connectors (push mode + acp1 + probel-sw02p…)

by-rune:
Replicates the cross-OS dhs_verbtest verb-test pattern (merged #619) to the remaining connectors. Each play deploys the per-OS binary + fixtures (control-node cross-build, no Go on target), serves/listens on the target, and drives EVERY verb asserting output — Linux (SSH) + win11 (WinRM). Batch 1 (this commit): adds a PUSH mode to the role (vt_mode serve|push; posix_push/windows_push) for tally/OSC protocols; probel-sw02p-verbs.yml (15 verbs); acp1-verbs.yml (manifest+DM deploy, TCP Mode B). Still to add on this branch: acp2, emberplus (serve), tsl, osc (push), cerebrum-nb (external) + test.yml wiring. Verbs cited to cmd/dhs source; authored to the proven WinRM/SSH deploy pattern; live run on the control node. Co-Authored-By Claude Opus 4.8.

### DIFF
--- a/ansible/playbooks/acp1-verbs.yml
+++ b/ansible/playbooks/acp1-verbs.yml
@@ -1,0 +1,102 @@
+---
+# ACP1 VERB-TEST contract play — deploy the per-OS binary + the committed
+# synapse manifest/DM fixture, serve our acp1 producer on the target, and drive
+# the consumer verbs against it end to end over the wire. Mirrors
+# probel-sw08p-verbs.yml via the dhs_verbtest role (cross-build on control node,
+# deploy per OS, serve, drive). Per ADR-0025 deliverable 3 + ADR-0016.
+#
+# ACP1 has NO single committed tree.json — the producer assembles its tree from
+# a manifest (--manifest) + the referenced DMs under <cache-dir>/dm/acp1/ per
+# ADR-0022 (same shape acp2-integration.yml uses). We therefore deploy the
+# manifest AND each DM to a cache layout under dhs_dir and point --manifest /
+# --cache-dir at the DEPLOYED paths (never the repo path).
+#
+# Serve over TCP (Mode B, --transport tcp): the role's wait_for probes a TCP
+# port, and the consumer dials the same with --transport tcp. (Mode A UDP can't
+# be wait_for-probed cross-OS; TCP gives a deterministic loopback contract.)
+#
+# The producer is ephemeral (TTL auto-expiry), so run-twice = same result.
+#
+#   ansible-playbook -i inventory/hosts.ini playbooks/acp1-verbs.yml \
+#     -e @secrets/credentials.json
+- name: ACP1 verb test (deploy binary + manifest/DM fixture, serve on target, drive verbs)
+  hosts: linux:windows
+  gather_facts: true
+  vars:
+    vt_host: "127.0.0.1"
+    vt_port: 12071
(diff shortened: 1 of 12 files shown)


by-rune:
Rebased on today's main (merge commit, one conflict: `ansible/playbooks/cerebrum-nb-verbs.yml`, resolved by keeping main's version). Every play the branch adds passes `ansible-playbook --syntax-check` on the control node. Left as draft because a live run is not green yet — two findings from running `tsl-verbs.yml` on the fleet (2026-09-17):

1. **The push mode's listener leaves no capture.** `Read the listener capture back from the target` fails on every Linux host: `file not found: /usr/local/bin/dhs-tsl-v50-listen.out`. The async `shell` that starts `dhs consumer tsl-v50 listen ... > capture` never produced the file, so either the async wrapper did not run it or the listener's argv is refused before the shell redirect matters. Worth a run with `poll` and the stderr kept instead of `2>/dev/null`.
2. **The role cross-builds on the play's host and deploys that build over `/usr/local/bin/dhs`** on every fleet host — the same path the `dhs_host` role now owns from a pinned release (v0.22.0 since #1076). The run replaced the release binary on four LXCs with a branch build; I restored them with `dhs_host`. This is inherited from the merged #619 role, not new here, but any verb play now fights the release pin: the role should deploy to its own path (as `snell-rollcall-ipshare.yml` does under `/opt/dhs/...`) or take a prebuilt artifact (`DHS_BIN`, as `snell-rollcall-integration.yml` does), since the control node has no Go toolchain (`[Errno 2] No such file or directory: b'go'` on dhs-debian).

by-rune:
**All seven verb plays now run green across the fleet.** Live runs from the dhs-tools runner against the whole inventory (dhs-debian, dhs-rocky, dhs-tools, dhs-ubuntu, win11):

- acp1, acp2, emberplus, probel-sw02p, probel-sw08p: `failed=0` on every host (serve-mode).
- tsl, osc: `failed=0` on the Linux hosts (push-mode; see below).
- The pinned release binary at `/usr/local/bin/dhs` (v0.23.1, dhs_host role) is untouched by every run — the role now deploys the build under test to `/opt/dhs-verbtest` (Linux) / `C:/dhs-verbtest` (Windows).

Fixes made to get here (findings #1 and #2 from my earlier comment resolved, plus arg reconciliation against the binary):
- Role deploys to its own dir, never over the release binary.
- Push listener redirect on one line so the capture exists while the listener runs (the empty-capture finding).
- Windows: read the capture with `ansible.builtin.slurp` (there is no `win_slurp`), stop the listener before reading, and free a locked `dhs.exe` before deploy.
- Fixture parent directories created before copy.
- Verb args reconciled to the binary: sw08p `salvo-connect --src`, `protect-name --device`, `update-name --first-id/--names/--type`, `watch --timeout`, `bench --size/--skip-warmup`; acp2 `diag` dropped (device-only, dials the ACP2 default port, not the ephemeral serve).

Known gap: tsl/osc **push capture on Windows** loses the PowerShell-buffered listener stdout on force-kill, and osc `listen` has no clean-exit flag to flush first. Scoped those two push plays to Linux and opened #1094 for the dhs-side `listen --duration/--count` fix. Idempotency (run-twice = 0 changes) verifying now.

PR #720, opened 2026-08-19
fix: root-cause CI timing flakes; ADR-0029 event-step matrix; drop retries + coverage union (#694)

### LINKED ISSUE #694: ci: full pipeline audit - one clear step chain per event type, root-cause flakes, restore speed
Owner-requested plan (2026-08-17, execution deferred): the CI grew by patching symptoms and drifted from the original clear/fast shape. Audit and redesign top-down, one documented step chain per event type: COMMIT (pre-commit local: vet+lint), PR (tests once, fast, no retries needed), MERGE to main (full verification per commit, never cancelled), RELEASE (release-please PR merge -> tag -> binaries; changelog owned by release-please only - stale block cleaned in #692/#693). Core work items: (1) ROOT-CAUSE the timing flakes (acp2/consumer TestRunDiagnostics_ConnectionClosed and the coverage-union class) so the 3-attempt retry loops and the 5-attempt coverage UNION can be REMOVED - the union was added for scheduling-dependent coverage lines, is slow, and can mask a real bug (owner flagged this twice); (2) speed budget: unit tests + compile were fast originally - measure per-job time, cut the coverage-floor sweep cost (single pass), consider paths-ignore for docs-only commits; (3) document the final event->steps matrix in an ADR so it cannot drift again (candidate ADR-0028); (4) revisit what #690/#691 layered on once flakes are root-caused (retries become dead code). Deliverable: ADR + slimmed ci.yml + flaky tests fixed deterministically.

by-rune:
Closes #694.

The owner-requested CI audit, delivered as its three work items: the timing flakes root-caused and fixed deterministically, the compensating retry/union machinery removed, and the pipeline written down as one auditable step chain per event type (ADR-0029).

## Root causes fixed

**acp2/consumer — `TestRunDiagnostics_ConnectionClosed` class** (the live 2026-08-17 rocky9 main-push red). Request waiters selected on `<-s.done` next to their reply channel. When the peer replied and hung up in one burst, both arms were ready and Go picks a ready select arm pseudo-randomly — so "reply" vs "connection closed" was a coin flip, and the statement coverage of each arm was scheduling-dependent (the reason the coverage union existed). Fix: a deterministic waiter-death protocol — the read loop's exit sweeps the waiter table under the registration mutex (nil sentinel to every waiter, an already-delivered reply survives), and late registrations fail fast. Every waiter receives exactly one value; no racy select arm remains. The diag probe/announce timings became injectable, so the four formerly multi-second diag tests now run in milliseconds with strengthened assertions.

**probel-sw02p/provider — `TestIntegrationSalvoBroadcastsConnectedAcrossSessions`.** The test grabbed a free port by listen-close-relisten; in a parallel sweep another process can steal the port inside that window. Fix: `ServeListener` exported on the concrete provider type (neutral interface untouched) so the test pre-binds `127.0.0.1:0` and hands the listener over — no rebind window, no dial-poll.

## CI slimmed (ADR-0029)

- Unit tests run **once** everywhere; both 3-attempt retry loops removed.
- Coverage floor computed from the single whole-repo `coverage.out` with awk — the 5-attempt per-package union sweep removed (up to 135 package test runs saved on the ubuntu job, expected ≈ 5.8 → ≈ 4 min). Missing-package hard-fail guard added; misses print exact uncovered blocks.
- `paths-ignore` for docs-only commits considered and rejected (documented in the ADR): every merge commit on main keeps its own complete verification and release-please bookkeeping.

## Evidence

- acp2/consumer: 100.0% coverage on 5 consecutive single runs (was a stable-but-racy 99.6/100 union split); 30× stress of the flake class green.
- probel-sw02p/provider: 10× stress green; 100.0% on 3 single runs.
- All 27 floor packages: 100.0% single-pass, one full sweep.
- New floor step mirrored locally against a real whole-repo profile: 27/27 pass, renamed-package guard fires, sub-floor path prints uncovered blocks.
- 3 consecutive full-suite runs green with retries already removed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,28 +46,12 @@ jobs:
       - name: Vet
         run: go vet ./...
 
+      # One pass, no retries (ADR-0029): the timing flakes that motivated
+      # the old 3-attempt loop were root-caused and fixed deterministically
+      # (acp2 waiter-death protocol, probel-sw02p pre-bound listener). A red
+      # here is a real problem in this commit.
       - name: Unit tests
-        shell: bash
-        run: |
-          set -o pipefail
-          # A few consumer/provider tests exercise goroutine / reconnect /
-          # timer timing (net.Pipe, fast reconnect backoff, keep-alive) that
-          # can flake on a loaded CI runner even when the code is correct —
-          # the same class the coverage-floor step already retries. Retry the
-          # suite up to 3x and pass on the first green; a genuine failure
-          # fails all three attempts. Root-causing individual flakes is still
-          # tracked separately — this only stops them gating merges.
-          for attempt in 1 2 3; do
-            echo "::group::unit tests (attempt $attempt/3)"
-            if go test -count=1 -race ./...; then
-              echo "::endgroup::"
-              exit 0
-            fi
-            echo "::endgroup::"
-            echo "::warning::unit tests failed on attempt $attempt/3; retrying"
-          done
-          echo "::error::unit tests failed after 3 attempts"
-          exit 1
+        run: go test -count=1 -race ./...
 
       - name: Unit tests (coverage)
         if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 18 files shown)


by-rune:
First CI run went red on the ubuntu leg — and that red is the new floor step working as designed: it named one residual scheduling-dependent line the old 5-attempt union had been silently absorbing:

```
internal/acp1/consumer: 99.9% (floor 100%, 1145 blocks)
uncovered blocks (cover these deterministically):
    dhs/internal/acp1/consumer/discover.go:98.22,100.5  (1 stmt)
```

Root cause: `Discover`'s receive loop had a loop-top `remaining <= 0 → return` that only ever executed when a datagram landed exactly at the discovery deadline — coverage by coincidence. Fixed in `54f5220` by deleting the racy branch: each `Receive` now runs under a ctx capped at the deadline, so deadline expiry always exits through the single err arm (the transport maps an expired read deadline to `context.DeadlineExceeded`). acp1/consumer holds 100.0% on 3 consecutive single-pass runs.

Exactly the failure mode ADR-0029 predicts: with the union gone, a coincidence-covered line fails loudly with its file:line instead of passing ~1−p⁵ of the time.

PR #828, opened 2026-08-23
feat(amwa/sdp): stdlib-only SDP parser for ST 2110 transport files (#827)

by-rune:
## What

New stdlib-only package `internal/amwa/codec/sdp` — a parser for the SDP transport files NMOS carries (IS-04 `manifest_href`, IS-05 `transport_file`).

## Why

The repo could not read an SDP at all. The only SDP code was `sdpFor()` in `internal/amwa/provider/node_endpoints.go:54`, which *emits* a fixed minimal audio body. Everything the wire actually carries was unreachable in code:

- the multicast address **per SMPTE 2022-7 leg** (`c=` inside each `m=`, resolved through `a=group:DUP` + `a=mid`);
- the PTP grandmaster and domain (`a=ts-refclk`) — the check that catches a sender quoting its own clock identity instead of the fabric grandmaster;
- the essence parameters (`a=rtpmap`, `a=fmtp`, `a=ptime`) and the source address (`a=source-filter`).

Closes #827

## Scope

- [x] osc / tsl / cerebrum-nb / amwa

## Reference controller

- [x] N/A (no protocol behaviour change — read-only codec, not wired to any surface yet)

## Type

- [x] feat — new feature

## Files changed

| File | New / Modified | Description |
|------|---------------|-------------|
| `internal/amwa/codec/sdp/sdp.go` | new | model: Session, Media, Connection, RTPMap, FMTP, TSRefClk, SourceFilter, Group, Leg |
| `internal/amwa/codec/sdp/parse.go` | new | RFC 4566 line parser + ST 2110 attribute decoding + Deviation reporting |
| `internal/amwa/codec/sdp/sdp_test.go` | new | the three real essence shapes, legs, deviations, empty/non-SDP |
| `internal/amwa/codec/sdp/edge_test.go` | new | edge cases: session-level `c=`, property attributes, non-PTP ts-refclk, bare fmtp flags, malformed lines |
| `internal/amwa/codec/sdp/testdata/*.sdp` | new | 5 fixtures |

## Design decisions

**`Parse` returns `(*Session, []Deviation, error)`.** A recoverable defect becomes a Deviation and parsing continues — the connector compliance posture, absorb and surface rather than patch. Only a document that is not SDP at all is an error (`ErrEmpty`, `ErrNoVersion`). A caller wanting strictness treats a non-empty deviation slice as failure.

[...]

PR #836, opened 2026-08-23
feat(nmos): offline compliance audit of a captured plant (#835)

### LINKED ISSUE #835: feat(nmos): offline compliance audit of a captured plant
## Problem

We can export a customer's NMOS plant, and then we read the result by eye. A 44-node registry export is ~1.5 GB of JSON across ~450 folders. Nobody finds a multicast collision in that by scrolling.

The export already proved it holds real defects that no single device reports:

- a registry listing **45 nodes at v1.1 and 39 at v1.3** — a controller on the highest minor silently sees six fewer devices, and nothing errors
- a device returning **502 on every `transportfile`** — the endpoint exists and is broken, which a controller reads as the device failing rather than as absent
- a node **registered but unreachable** — controllers keep offering a device that is gone

Each of those is a finding an operator would take to the vendor. Today none of them is produced by a tool.

## What to build

`dhs consumer nmos audit --dir <export>` — offline, read-only, deterministic.

Offline is the point. The export happens once, on the customer's site, on their network. The audit then runs anywhere, any number of times, and reaches the same verdict every run — so the report can be attached to a ticket, and re-run six months later against the same export to prove what changed.

### Checks

**Per device**
| Area | What it catches |
|---|---|
| API surface | what each device actually exposes, per API per minor |
| IS-05 absent | a node publishing senders that no controller can ever route |
| `controls[]` cross-check | an advertised control whose API does not answer, and an API no control advertises — a spec-following controller walks `controls[]` and nothing else |
| `resource_core` | non-UUID ids, non-TAI versions, duplicate ids |
| reference graph | Sender→Flow→Source→Device→Node edges pointing at resources that are not there |
| `manifest_href` | null before v1.3, null on a subscribed sender, empty string |
| BCP-002-01 | missing or malformed group hints — without them a controller cannot group or break away essences |
| BCP-004-01 | receivers with no `caps.media_types`; `constraint_sets` with no `caps.version` |
| IS-05 active | master-enabled with no destination, port 0, every leg `rtp_enabled=false` |
| ST 2022-7 | two NICs bound but one leg staged; both legs on one `/24` |
| transport report | 5xx on spec-defined endpoints, unreachable endpoints, paging defects |

**Plant-wide** — these three do not exist at device scope:
- two senders on the same multicast group and port
- version isolation hiding resources from a highest-minor controller
- a node the registry lists that answers nowhere

### Output

`text` for reading, `json` for tooling, `jsonl` for line-diffing two exports. `--min-severity` to filter, `--fail-on` to gate CI. Without `--fail-on` the verb exits 0 — reporting is not the same as failing, and each CI job decides its own threshold.

Every finding carries the clause it rests on, so the report goes to the vendor as-is.

## Out of scope here

- SDP correlation (sender SDP vs IS-05 active destination) — needs the SDP codec (#828) merged first
- live probing — a separate verb, tracked separately

## Definition of done

- [ ] package builds, `go vet` clean, `golangci-lint` 0 issues
- [ ] fixture plant committed, using documentation address ranges only
- [ ] coverage >= 90%
- [ ] two runs over the same capture produce byte-identical reports
- [ ] `dhs consumer nmos audit --help` documents every flag

by-rune:
## What

`dhs consumer nmos audit --dir <export>` — an offline compliance audit of a captured NMOS plant.

## Why

We could already export a customer's plant. Reading the result was manual: a 44-node registry export is roughly 450 folders of JSON, and nobody finds a multicast collision by scrolling it.

The export already held real defects, and none of them is reported by the device that has them:

| Observed on a live plant | What a controller sees |
|---|---|
| registry lists 45 nodes at v1.1, 39 at v1.3 | six devices silently absent, no error anywhere |
| 502 on every `transportfile` | the device failing, not an endpoint that is missing |
| node registered, answers nowhere | a device it keeps offering to operators |

Closes #835

## Scope

## Reference controller

- [x] N/A (offline analysis; nothing on the wire changes)

## Type

- [x] feat — new feature

## Files changed

| `internal/amwa/audit/finding.go` | new | Finding + Severity, stable ordering |
| `internal/amwa/audit/harvest.go` | new | export reader; registry + followed nodes as one tree |
| `internal/amwa/audit/audit.go` | new | check registry, inventory, Run |
| `internal/amwa/audit/checks_is04.go` | new | API surface, `controls[]`, resource_core, reference graph, manifest, BCP-002-01, BCP-004-01 |
| `internal/amwa/audit/checks_is05.go` | new | active endpoints, transport params, ST 2022-7 leg structure |
| `internal/amwa/audit/checks_report.go` | new | HTTP faults, paging defects, version isolation |
| `internal/amwa/audit/checks_plant.go` | new | multicast collisions, registration state |
| `internal/amwa/audit/render.go` | new | text / json / jsonl |
| `internal/amwa/audit/testdata/plant/` | new | a fixture registry + two followed nodes |
| `cmd/dhs/cmd_nmos_audit.go` | new | the verb |

## Design decisions

**Offline, and deterministic.** The export happens once, on the customer's site, on their network. The audit then runs anywhere, any number of times, over those bytes. A test asserts two runs produce byte-identical reports — without that, diffing this month's audit against last month's means nothing.

**Three checks are plant-wide, and cannot be anything else.** A multicast collision is invisible from inside either device: both are configured correctly and both report healthy. Version isolation is invisible from inside the registry: it is answering every query correctly. An unreachable registered node is invisible from the node, which is not running. These only appear when the whole export is read at once, which is exactly what a registry export gives you.

**Node APIs are read at the highest minor; a registry's query API at every minor.** A node repeats the same resources at each minor it serves, so collapsing is safe and saves 4× the requests. A registry does not: IS-04 version isolation means a v1.3 query hides what was registered at v1.1. Taking the highest there would have reported 39 nodes on a 45-node plant.

**Absence of a collection is not evidence that a reference dangles.** A partial export would otherwise flood with false dangling-reference findings. The graph check stays silent when the target collection was never captured.

**Faults group by endpoint shape, not URL.** A device that 502s on all 176 of its senders produces one finding naming 176. Ungrouped, the report is unreadable at exactly the moment it matters most.

**`--fail-on` is opt-in.** Without it the verb exits 0 even having found problems. Reporting is not the same as failing, and each CI job picks its own threshold.

## Test results

| | |
|---|---|
| `go test ./internal/amwa/audit/...` | pass |
| Coverage | **96.5% of statements** |

Run against the committed fixture plant:

```text
SUMMARY

  CRITICAL   3
  ERROR      9
  WARN       7
  INFO       6
```

One of the three criticals, as it prints:

```text
NMOS-PLANT-MCAST-COLLISION     2 senders emit to 233.252.0.20:20000:
                               cam-01 sender/4444...444 leg0,
                               mv-01 sender/bbbb...bbb leg0
                                 spec SMPTE ST 2110-10 / IS-05 v1.1 §4.2 destination_ip
                                 → receivers joining this group get interleaved essence from both sources
```

## Device tested

- [x] No device needed — the fixture plant is committed, and uses the 233.252.0.0/24 and 198.51.100.0/24 documentation ranges. No customer addressing is in this PR.

## Not in this unit

- **SDP correlation** — comparing each sender's published SDP against its IS-05 active `destination_ip` is the strongest cross-check available, and it needs the SDP codec (#828) merged first. Follow-up.
- **Live probing.** This verb reads bytes; connecting to a registry and running assertions is its own verb.

## Checklist

- [x] `go vet ./...` clean
- [x] Integration tested on VM before PR — N/A; the verb has no network path

## Approval

@yboujraf — requesting review


by-rune:
Superseded by the conformance-first approach: plant auditing now runs via the AMWA Testing tool suites against the live plant instead of offline capture audits. Issue #835 stays open for re-triage. Branch kept.

PR #838, opened 2026-08-23
feat(nmos): consumer nmos export - plant capture in Go (#837)

### LINKED ISSUE #837: feat(nmos): consumer nmos export - plant capture in Go, replacing the PowerShell exporter
## Problem

The plant exporter is a PowerShell script. That is fine for an operator sitting at a Windows box, and wrong for everything else:

- it does not run on the Linux control node, or in CI, or in a container
- it is not covered by any test, so a regression in it is found by an operator on a customer site
- the audit (#835) reads its output format, and nothing enforces that the two stay in step
- an operator has to be handed a second tool alongside `dhs`

## What to build

`dhs consumer nmos export --target <host:port> --out <dir>` — the same capture, in Go, in the same on-disk layout the audit already reads.

```
<out>/<host_port>[__label]/
  device.json     identity, written at start and again at the end
  tree.json       every resource captured, keyed api -> version -> resource
  report.txt      one line per request: status, path, and the notes the audit parses
  raw/            each response body as fetched
  sdp/            every SDP recovered, per sender and per subscribed receiver
  nodes/          for a registry: each node it lists, followed and captured the same way
```

### Behaviour that is not optional

These are the lessons the PowerShell version learned the hard way, and the Go version has to carry them or it captures less than the script it replaces:

| Rule | Why |
|---|---|
| Walk a **registry's query API at every minor** | IS-04 version isolation: v1.1 returned 45 nodes where v1.3 returned 39. Collapsing to the highest minor loses six devices. |
| Walk a **node's own API at the highest minor only** | A node repeats the same resources at every minor. Walking all four quadruples the request count for identical bytes. |
| Follow `Link: rel="next"` | Without paging, a 68-node registry captured 11 nodes. |
| Stop when the paging cursor does not advance | Otherwise a registry with a stuck cursor loops forever. Record it — that is a finding. |
| Follow `manifest_href` **verbatim** | IS-04 puts the SDP where the sender says it is. Guessing `/transportfile` produced 1088 errors against a real device. |
| Request `transporttype` only on IS-05 >= v1.1 | It does not exist at v1.0. One live run produced 5,951 404s. |
| Never create `sdp/` or `raw/` at the output root | Every folder must be attributable to a device. |

### CLI

```
--target host:port     device or registry to capture
--out DIR              output directory (default ./nmos-export)
--https                use TLS
--deep                 also fetch staged / constraints / transporttype per endpoint
--all-versions         walk every minor of every API, not just where it matters
--max-nodes N          cap how many registered nodes to follow
--no-sdp               skip SDP retrieval
--timeout DURATION     per-request deadline
```

## Definition of done

- [ ] output is byte-compatible with what `dhs consumer nmos audit` reads
- [ ] round-trip test: httptest registry + node -> export -> audit, asserting the audit finds the defects the fake plant was built with
- [ ] paging, version isolation and `manifest_href` behaviours each have their own test
- [ ] `go vet` + `golangci-lint` clean, coverage >= 90%
- [ ] runs on Linux and Windows

## Follow-on

`dhs consumer nmos probe` — live protocol conformance (version negotiation, unknown-version handling, CORS, content types, paging semantics, heartbeat lifecycle) with per-endpoint latency, emitted as JSONL. Filed separately once this lands.

by-rune:
> **Stacked on #836.** Base is `feat/nmos-audit`; GitHub retargets to `main` when #836 merges. Review #836 first.

## What

`dhs consumer nmos export --target <host:port>` — plant capture in Go, writing exactly the layout `dhs consumer nmos audit` reads.

Together the two verbs are one workflow: **capture on the customer's network, audit anywhere.**

```bash
dhs consumer nmos export --target 10.44.55.56:8080 --out ./plant
dhs consumer nmos audit  --dir ./plant --min-severity warn
```

## Why

The exporter was PowerShell. That is fine at a Windows desk and wrong everywhere else: it does not run on the Linux control node, in CI, or in a container; no test covered it; and nothing kept its output format in step with the audit that reads it.

Closes #837

## Scope

## Reference controller

- [x] N/A — read-only capture, no wire behaviour changes

## Type

- [x] feat — new feature

## Files changed

| `internal/amwa/session/export/export.go` | new | the capture: API discovery, per-spec walks, paging, SDP, node following |
| `internal/amwa/session/export/export_test.go` | new | the behaviours that cost a real capture when missing |
| `internal/amwa/session/export/export_edge_test.go` | new | generic API walks, TLS, cancellation, malformed answers |
| `cmd/dhs/cmd_nmos_export.go` | new | the verb |
| `cmd/dhs/cmd_nmos_export_audit_test.go` | new | export → audit round-trip over a fake plant |

## Design decisions

**The capture never repairs.** A 502, a stuck cursor, an unreachable node — each is recorded and the walk continues. The audit turns those recordings into findings, and it can only do that if the capture did not paper over them first.

**Four behaviours carried over from the script**, each of which cost a real capture when it was missing:

| Rule | What it cost |
|---|---|
| Walk a registry's query API at **every** minor | version isolation hid 6 of 45 nodes |
| Follow `Link: rel="next"` | a 68-node registry captured 11 |
| Follow `manifest_href` **verbatim**, relative form included | a constructed `/transportfile` produced 1088 errors on a device that serves SDP elsewhere |
| Request `transporttype` only on IS-05 ≥ v1.1 | one run produced 5,951 404s |

**A node's own API is walked at the highest minor only.** It repeats identical resources at every minor; walking all four quadruples the requests for the same bytes. The asymmetry with the query API is the whole point — the two are different for a reason, and treating them alike loses either devices or time.

## Two defects the round-trip test found

Neither was visible from either package alone.

**A node that answered nothing was left as an empty harvest folder.** It then audited as a fourth device with no APIs, and the registry advertising a dead node produced no finding at all. It is now removed and recorded as `SKIP` on the parent — which is exactly the line the audit reads to report `NMOS-NODE-UNREACHABLE`.

**A collection captured as `[]` was treated like one never captured.** The audit suppresses dangling-reference findings when a collection was not reached, to avoid flooding a partial export with noise. But `"flows": []` is the device *saying it has no flows*, so a sender pointing at one is broken. Conflating the two silently suppressed the finding on every node that publishes empty collections — which is most of them. Fixed in #836, with its own test.

## Test results

| | |
|---|---|
| `go test ./internal/amwa/session/export/...` | pass |
| Coverage | **91.8% of statements** |

The round-trip test builds a fake plant carrying seven known defects, exports it, audits it, and asserts every one is found:

```text
NMOS-PLANT-MCAST-COLLISION      both nodes on 233.252.0.20:20000
NMOS-QUERY-VERSION-ISOLATION    3 nodes at v1.1, 1 at v1.3
NMOS-NODE-UNREACHABLE           a registered node that answers nowhere
NMOS-HTTP-SERVER-FAULT          502 on the transport files
NMOS-IS04-DANGLING-REF          sender points at an unpublished flow
NMOS-IS04-CONTROL-MISSING       IS-05 served, no sr-ctrl control
NMOS-2022-7-SINGLE-LEG          two NICs bound, one leg staged
```

A rename on either side of the file format now fails the build, instead of silently producing empty audits.

## Device tested

- [x] No device needed — `httptest` throughout. This is what makes the workflow runnable in CI; it does not replace running it against the real plant, which is the next step.

## Not in this unit

`dhs consumer nmos probe` — live protocol conformance (version negotiation, unknown-version handling, CORS, content types, heartbeat lifecycle) with per-endpoint latency, emitted as JSONL. That is the AMWA-test-tool-shaped verb, and it belongs on top of this one.

## Checklist

- [x] `go vet ./...` clean
- [x] Integration tested on VM before PR — httptest plant; real-plant run pending

## Approval

@yboujraf — requesting review

### DIFF
--- a/cmd/dhs/cmd_nmos.go
+++ b/cmd/dhs/cmd_nmos.go
@@ -50,10 +50,12 @@ func runNMOSConsumer(ctx context.Context, args []string) error {
 		return runNMOSWalk(ctx, rest)
 	case "events":
 		return runNMOSEventsConsumer(ctx, rest)
+	case "export":
+		return runNMOSExport(ctx, rest)
 	case "audit":
 		return runNMOSAudit(ctx, rest)
 	}
-	return fmt.Errorf("consumer nmos: unknown verb %q (expected: discover, system, walk, events, audit)", verb)
+	return fmt.Errorf("consumer nmos: unknown verb %q (expected: discover, system, walk, events, export, audit)", verb)
 }
 
 // runNMOSProducer dispatches `dhs producer nmos <verb> [args]`.
@@ -576,6 +578,29 @@ JSON Schema, prints the parsed fields.
   --direct host:port    Skip discovery; fetch /global directly from this peer
   --api-ver V           Requested IS-09 version (default v1.0)
   --api-proto P         Requested protocol filter (default http)
+
+export — capture a device, or a registry and every node it lists, into the
+layout the audit verb reads. Failures are recorded, never worked around: a
+502, a stuck paging cursor or an unreachable node all land in report.txt and
+the walk continues.
+  --target host:port    Device or registry to capture (required)
+  --out DIR             Output directory (default nmos-export)
+  --https               Use TLS
+  --deep                Also fetch staged / constraints / transporttype per IS-05 endpoint
+  --all-versions        Walk every minor of every API
+  --no-sdp              Skip SDP retrieval
+  --raw                 Also keep every verbatim response body under raw/
+                        (tree.json already holds them; raw/ was 14,719 of
+                        27,299 files on a real plant, and every path that
+                        exceeded the Windows 260-character limit)
+  --max-nodes N         Cap how many registered nodes to follow (0 = no cap)
+  --workers N           Capture N nodes at once (default 6)
+  --page-limit N        Resources per Query API page (default 100). A registry
+                        may clamp it; the applied value is on the PAGING line
+                        in report.txt
+  --timeout DURATION    Per-request deadline (default 10s)
+  --no-stamp            Name folders without a timestamp
+  --quiet / --json      Suppress progress / print the summary as JSON
 
 audit — compliance-audit a captured plant, offline. Reads an export directory
 (device.json + tree.json + report.txt + raw/ + sdp/ per device, nested under
--- a/cmd/dhs/cmd_nmos_export.go
+++ b/cmd/dhs/cmd_nmos_export.go
@@ -1,0 +1,120 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"dhs/internal/amwa/session/export"
+)
+
+// runNMOSExport implements `dhs consumer nmos export`.
+//
+// It captures a device — or a registry together with every node it
+// lists — into the on-disk layout `dhs consumer nmos audit` reads. The
+// two verbs are one workflow: capture on the customer's network, audit
+// anywhere.
+//
+// The capture never repairs what it finds. A 502, a stuck paging
+// cursor, a node that answers nowhere: each is recorded and the walk
+// continues. Those recordings are what the audit turns into findings,
+// and it can only do that if the capture did not quietly paper over
+// them first.
+func runNMOSExport(ctx context.Context, args []string) error {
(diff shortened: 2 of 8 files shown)


by-rune:
Superseded: plant capture/export re-scoped by the rework; the walk/export verbs will be documented + finalized in the post-spec-100 CLI unit. Issue #837 stays open for re-triage. Branch kept.

Issue #841, opened 2026-08-23
feat(nmos): IS-05 Connection API server on producer nmos serve


by-rune:
## Problem

Our Node serves IS-04 and nothing else. A controller can discover it, list its senders, and then do nothing with them: IS-05 is the only standard way to set a sender''s destination, and we do not serve the API.

The audit reports this against our own node, as CRITICAL:

```
NMOS-IS05-ABSENT   node publishes 1 sender(s) but serves no /x-nmos/connection API
                     spec IS-05 v1.1 §4 Connection API
                     -> the node is read-only to any controller; routing it needs a vendor-specific path
```

This is the gap that stops the fleet demo short of an actual routing change. We can show discovery, registration, grains over the Query WebSocket, and a controller that sees everything — and then cannot move a single stream.

## What to build

The IS-05 Connection API server on `dhs producer nmos serve`, at v1.0 and v1.1:

| Endpoint | Method | Notes |
|---|---|---|
| `/single/senders` · `/single/receivers` | GET | id lists |
| `/single/{side}/{id}/constraints` | GET | one entry per leg |
| `/single/{side}/{id}/staged` | GET, PATCH | the write path |
| `/single/{side}/{id}/active` | GET | read-only, per spec |
| `/single/{side}/{id}/transporttype` | GET | v1.1+ only |
| `/single/senders/{id}/transportfile` | GET | `application/sdp`, generated from the active params |
| `/bulk/senders` · `/bulk/receivers` | POST | |

### Behaviour the client work already pinned

The IS-05 client (#831) established the wire shape from real hardware, and the server has to be the other half of it:

- `transport_params` is a **partial** update. An absent field means "leave it"; `0` means zero. The staged object merges rather than replaces.
- All three activation modes: `activate_immediate`, `activate_scheduled_relative`, `activate_scheduled_absolute`.
- `master_enable` gates the sender independently of what is staged.
- The transport file is **generated**, not stored. Set `destination_ip` / `destination_port`, activate, and the SDP the node serves changes to match — that is the whole mechanism by which IS-05 changes what a sender emits.
- Activation must bump the IS-04 sender''s `version` and re-register, so the registry and every subscribed controller see the change.

### The check that closes the loop

Once this lands, the end-to-end assertion is possible for the first time:

1. `dhs consumer nmos connect` PATCHes staged with a new multicast group and activates
2. the node regenerates its SDP
3. the node re-registers, the registry updates, the Query WebSocket emits a `modified` grain
[...]

Issue #840, opened 2026-08-23
fix(nmos/registry): version root 404s without a trailing slash


by-rune:
## Found by

`dhs consumer nmos probe` (#839), run against our own registry. This is the first defect the probe found, and it found it in us.

```
WARN  PROBE-PATH-001  a version root answers with and without a trailing slash
        /x-nmos/query/v1.3/ -> 200 but /x-nmos/query/v1.3 -> 404
        spec IS-04 v1.3 §4 API paths
```

Reproduced by hand against `dhs registry nmos serve --bind :18235`:

| Request | Result |
|---|---|
| `GET /x-nmos/query/v1.3/` | 200, lists `nodes/ devices/ …` |
| `GET /x-nmos/query/v1.3` | 404 |

## Why it matters

Controllers build these URLs by concatenation and disagree about the trailing slash. Serving one and 404ing the other works with whichever half of the market happens to add the slash.

It is not a hard spec violation — IS-04 documents the paths with trailing slashes, and a client that follows them literally is fine. But the AMWA reference implementation answers both, so every peer tested against nmos-cpp will have been written assuming both work. That makes this an interop defect regardless of how the clause is read.

## Fix

Serve the version root at both spellings. Either register both patterns, or redirect the un-slashed form with a 301 — a redirect is arguably better, since it tells a client which form is canonical instead of quietly accepting either.

Applies to every API our registry and node mount, not just `query`.

## Definition of done

- [ ] `GET /x-nmos/<api>/<ver>` and `GET /x-nmos/<api>/<ver>/` both succeed, on registry and node
- [ ] a unit test per API asserting both forms
- [ ] `dhs consumer nmos probe --target <our registry>` reports PROBE-PATH-001 as PASS

by-rune:
Verified fixed on the live registry (9ace69d): GET without trailing slash answers 200 on /x-nmos/query, /x-nmos/query/v1.3 and /x-nmos/registration/v1.3. IS-04-02 auto_query/auto_reg tests all pass.

Issue #173, opened 2026-04-30
nmos: AMWA NMOS Testing harness wiring — Docker + isolated bridge + per-suite digest pinning


by-rune:
Wire the AMWA NMOS Testing tool (Docker, github.com/AMWA-TV/nmos-testing) into tests/integration/nmos/ — one suite per spec / version. Per internal/amwa/docs/conformance.md: isolated Docker bridge (NEVER network_mode: host), image-digest pinned, Pass/Fail/CouldNotTest gating in CI. Sub-tracker of #146.

by-rune:
Superseded/delivered in altered form: the AMWA Testing tool runs on dhs-tools (Docker) against the Ansible-managed plant, driven by the sweep runners; per-spec Ansible validation roles are the next program unit. The isolated-bridge + digest-pinning design of this issue is no longer the plan.

PR #842, opened 2026-08-23
feat(nmos): consumer nmos profile - live conformance + latency, as JSONL (#839)

### LINKED ISSUE #839: feat(nmos): consumer nmos probe - live protocol conformance + latency, as JSONL
## Problem

The audit (#835) reads bytes. It is the right tool for "what does this plant look like, and where does it deviate" — and it is blind to everything that only exists in a live exchange:

- does an **unknown API version** 404, or does the device happily serve it?
- is **CORS** present? IS-04 requires it on every API, and without it no browser-based controller works at all
- is the **Content-Type** actually `application/json`? `application/sdp` on a transport file?
- does `paging.limit` get **honoured**, **clamped**, and **reported back** in the paging headers?
- does **`query.downgrade`** work, and does it actually return the resources version isolation hides?
- does a heartbeat for an **unregistered node** 404, so a node knows to re-register?
- **how slow is it?** A registry that answers correctly in 4 seconds fails a plant just as surely as one that answers wrongly.

None of that survives into an export, because an export records answers, not the conversation.

## What to build

`dhs consumer nmos probe --target <host:port>` — connect, assert, report.

```bash
dhs consumer nmos probe --target 10.44.55.56:8080 --format jsonl --out results.jsonl
```

One JSONL record per assertion:

```json
{"id":"PROBE-VER-002","name":"unknown API version is rejected","spec":"IS-04 v1.3 §6.1",
 "target":"10.44.55.56:8080","status":"FAIL","detail":"GET /x-nmos/query/v9.9/nodes returned 200, want 404",
 "duration_ms":12.4}
```

JSONL because it appends across runs and diffs as plain lines, so "did this device get better or worse since the firmware update" is a `diff`, not a reading exercise.

### Metrics

Every request is timed, and the run ends with per-endpoint-class latency (p50/p95/p99), request counts and failure counts. This is the measurement the optimisation work needs, and the number to put beside an incumbent product when the question is whether ours can replace it.

### Relationship to the AMWA Testing tool

This is not a replacement for it, and should not pretend to be. The AMWA suite is the reference and stays the reference (#173 / #183).

What this adds is a probe that runs **in one binary, with no Docker, against a production plant, in seconds** — the thing you can run on a customer's site during a commissioning window. The AMWA suite is what you run before you claim conformance. Both emit into the same JSONL shape so one report can carry both.

## Definition of done

- [ ] a check registry, so adding an assertion is one function
- [ ] `text` / `json` / `jsonl` output; `--fail-on` gates CI
- [ ] per-endpoint latency percentiles in the summary
- [ ] every check tested against an `httptest` device that gets it wrong, and one that gets it right
- [ ] `go vet` + `golangci-lint` clean, coverage >= 90%
- [ ] read-only: the probe never PATCHes, never activates, never registers anything

## Explicitly out of scope

**Any write.** A probe that stages a connection to test IS-05 is a probe that can take a live source off air. Write-path conformance needs a lab device and its own opt-in verb.

by-rune:
> **Stacked on #838**, which is stacked on #836. Base is `feat/nmos-export`; GitHub retargets as each merges. Review order: #836 → #838 → this.

## What

`dhs consumer nmos probe --target <host:port>` — live protocol conformance against one device, with per-endpoint latency, emitted as JSONL.

```bash
dhs consumer nmos probe --target 10.44.55.56:8080 --format jsonl --out results.jsonl
```

## Why

The audit reads bytes. A probe reads the conversation. Neither substitutes for the other:

| Question | Answered by |
|---|---|
| what does this plant look like, and where does it deviate | `audit` |
| does an unknown API version 404, or does the device serve it anyway | **probe** |
| is CORS present — without it no browser-based controller works at all | **probe** |
| is `paging.limit` honoured, clamped, and reported back | **probe** |
| does `query.downgrade` actually escape version isolation | **probe** |
| does health for an unregistered node 404, so the node knows to re-register | **probe** |
| how slow is it | **probe** |

Closes #839

## It found defects in us before it found any elsewhere

Run against our own registry and node:

| Finding | Filed |
|---|---|
| version root 404s without a trailing slash | #840 |
| our Node serves no Connection API — discoverable, unroutable | #841 |

## Scope

## Reference controller

- [x] N/A — read-only client

## Type

- [x] feat — new feature

## Files changed

| `internal/amwa/probe/probe.go` | new | session, transport, latency recording |
| `internal/amwa/probe/checks.go` | new | the thirteen assertions |
| `internal/amwa/probe/render.go` | new | text / json / jsonl |
| `internal/amwa/probe/probe_test.go` | new | a compliant device, then one broken rule at a time |
| `cmd/dhs/cmd_nmos_probe.go` | new | the verb |

## Design decisions

**Read-only by construction.** It never PATCHes, activates, or registers. A conformance tool that stages a connection is one that can take a live source off air, and this is meant to be safe to point at a plant that is on. Write-path conformance needs a lab device and its own opt-in verb.

**SKIP is reported, never folded into PASS.** A device with no Connection API skips the IS-05 assertions; that is not the same as passing them, and a green run must not hide an untested surface.

**Percentiles, not a mean.** A registry that answers in 8 ms and occasionally in 4 seconds has a fine mean and is unusable. Per-resource URLs collapse into one endpoint class, so a 176-sender device produces one row rather than 176.

**JSONL because it diffs.** One object per line, appendable across runs. "Did this device get better or worse since the firmware update" becomes a line diff. Latency rows are tagged `"kind":"latency"` so a consumer can tell them from results.

**This is not the AMWA Testing tool, and does not pretend to be.** That suite is the reference and stays the reference (#173/#183). This is what runs in one binary, no Docker, against a production plant, inside a commissioning window. Both emit the same JSONL shape, so one report can carry both.

## Test results

| | |
|---|---|
| `go test ./internal/amwa/probe/...` | pass |
| Coverage | **91.7% of statements** |

Every assertion is tested twice: against a device that gets it right, and one that breaks exactly that rule. The compliant baseline asserts *zero* failures, so a new check that misfires fails the build.

Against our own live registry:

```text
SUMMARY

  FAIL   0
  WARN   1     <- the trailing-slash defect, now #840
  SKIP   2
  PASS  13

LATENCY (ms)
ENDPOINT                                        REQS      P50      P95      P99   ERRS
/x-nmos                                            4      0.0      3.3      3.3      0
/x-nmos/query/v1.3/nodes                           3      0.5      1.1      1.1      0
```

## Device tested

- [x] Tested against a live `dhs registry nmos serve` + `dhs producer nmos serve` pair on this host, plus `httptest` devices for every failure mode. Not yet run against third-party hardware — that is the next step and needs the fleet.

## Not in this unit

- **IS-07 / IS-08 / IS-12 assertions.** The framework takes them as one function each; the specs land first.
- **Write-path conformance.** Deliberately excluded, see above.

## Checklist

- [x] `go vet ./...` clean
- [x] Integration tested before PR — against our own live registry and node

## Approval

@yboujraf — requesting review

### DIFF
--- a/cmd/dhs/cmd_nmos.go
+++ b/cmd/dhs/cmd_nmos.go
@@ -50,10 +50,12 @@ func runNMOSConsumer(ctx context.Context, args []string) error {
 		return runNMOSWalk(ctx, rest)
 	case "events":
 		return runNMOSEventsConsumer(ctx, rest)
+	case "export":
+		return runNMOSExport(ctx, rest)
 	case "audit":
 		return runNMOSAudit(ctx, rest)
 	}
-	return fmt.Errorf("consumer nmos: unknown verb %q (expected: discover, system, walk, events, audit)", verb)
+	return fmt.Errorf("consumer nmos: unknown verb %q (expected: discover, system, walk, events, export, audit)", verb)
 }
 
 // runNMOSProducer dispatches `dhs producer nmos <verb> [args]`.
@@ -576,6 +578,21 @@ JSON Schema, prints the parsed fields.
   --direct host:port    Skip discovery; fetch /global directly from this peer
   --api-ver V           Requested IS-09 version (default v1.0)
   --api-proto P         Requested protocol filter (default http)
+
+export — capture a device, or a registry and every node it lists, into the
+layout the audit verb reads. Failures are recorded, never worked around: a
+502, a stuck paging cursor or an unreachable node all land in report.txt and
+the walk continues.
+  --target host:port    Device or registry to capture (required)
+  --out DIR             Output directory (default nmos-export)
+  --https               Use TLS
+  --deep                Also fetch staged / constraints / transporttype per IS-05 endpoint
+  --all-versions        Walk every minor of every API
+  --no-sdp              Skip SDP retrieval
+  --max-nodes N         Cap how many registered nodes to follow (0 = no cap)
+  --timeout DURATION    Per-request deadline (default 10s)
+  --no-stamp            Name folders without a timestamp
+  --quiet / --json      Suppress progress / print the summary as JSON
 
 audit — compliance-audit a captured plant, offline. Reads an export directory
 (device.json + tree.json + report.txt + raw/ + sdp/ per device, nested under
--- a/cmd/dhs/cmd_nmos_export.go
+++ b/cmd/dhs/cmd_nmos_export.go
@@ -1,0 +1,114 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"dhs/internal/amwa/session/export"
+)
+
+// runNMOSExport implements `dhs consumer nmos export`.
+//
+// It captures a device — or a registry together with every node it
+// lists — into the on-disk layout `dhs consumer nmos audit` reads. The
+// two verbs are one workflow: capture on the customer's network, audit
+// anywhere.
+//
+// The capture never repairs what it finds. A 502, a stuck paging
+// cursor, a node that answers nowhere: each is recorded and the walk
+// continues. Those recordings are what the audit turns into findings,
+// and it can only do that if the capture did not quietly paper over
+// them first.
+func runNMOSExport(ctx context.Context, args []string) error {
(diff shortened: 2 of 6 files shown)


by-rune:
Superseded by #854 — this PR was auto-closed by GitHub when a bad rebase on my side briefly emptied its branch. Same unit, same content, new number.

Issue #847, opened 2026-08-24
fix(nmos): harvester drops single-element collections; track the script


by-rune:
## Problem

The operator-side harvester loses data, and it lived only in `captures/`, which is **gitignored** — so the fix could not be shared and the tool could not be reviewed or diffed.

### The data loss

A PowerShell function returning a **one-element array unrolls it to the bare element**. `Get-Json` did:

```powershell
return ($txt | ConvertFrom-Json)
```

So a node with exactly one Device recorded `"devices": {…}` instead of `[{…}]`.

Read strictly, that node then has *no* devices — and every Sender, Flow and Receiver on it points at a device that is not in the capture. On one real 7-node capture that produced **18,104 false dangling-reference findings**.

The same trap applies to `$x = if (…) {…} else {…}`: a statement block's output unrolls exactly like `return` does.

## Fix

Force-array at every assignment that stores a collection, with `/self` explicitly excluded because it is a single object:

```powershell
$got = Get-Json "$base/$res"
if ($res -eq 'self') { $bucket[$res] = $got } else { $bucket[$res] = @($got) }
```

Plus `@()` on `$tree.apis['_root']`, `versions`, `single_$side`, and `@(Get-JsonPaged …)`.

## Also: track it

Moved to `internal/amwa/assets/nmos-harvest.ps1`, alongside the other vendor tools for this protocol. It is an operator capture tool, not a provisioning play, so it does not conflict with the Ansible-only posture — but it **is** superseded by `dhs consumer nmos export` (#837), which cannot produce this class of bug because Go's `json.Marshal` always emits an array for a slice.

## Verification

Against a live registry + node pair:

| | before | after |
|---|---|---|
| `"devices"` starts with | `{` | `[` |
| `"senders"` starts with | `{` | `[` |
| `"self"` starts with | `{` | `{` (unchanged, correct) |
| audit dangling refs | 18,104 (plant) | **0** |

## Definition of done

- [x] tracked under `internal/amwa/assets/`
- [x] verified live: arrays survive, `/self` still an object
- [x] audit reports zero dangling references on the re-captured plant
- [ ] removed once `dhs consumer nmos export` is proven on a real plant

by-rune:
[...]

PR #848, opened 2026-08-24
fix(nmos): harvester drops single-element collections; track the script (#847)

by-rune:
## Problem

The operator-side harvester loses data, and it lived only in `captures/`, which is **gitignored** — so the fix could not be shared and the tool could not be reviewed or diffed.

### The data loss

A PowerShell function returning a **one-element array unrolls it to the bare element**. `Get-Json` did:

```powershell
return ($txt | ConvertFrom-Json)
```

So a node with exactly one Device recorded `"devices": {…}` instead of `[{…}]`.

Read strictly, that node then has *no* devices — and every Sender, Flow and Receiver on it points at a device that is not in the capture. On one real 7-node capture that produced **18,104 false dangling-reference findings**.

The same trap applies to `$x = if (…) {…} else {…}`: a statement block's output unrolls exactly like `return` does.

## Fix

Force-array at every assignment that stores a collection, with `/self` explicitly excluded because it is a single object:

```powershell
$got = Get-Json "$base/$res"
if ($res -eq 'self') { $bucket[$res] = $got } else { $bucket[$res] = @($got) }
```

Plus `@()` on `$tree.apis['_root']`, `versions`, `single_$side`, and `@(Get-JsonPaged …)`.

## Also: track it

Moved to `internal/amwa/assets/nmos-harvest.ps1`, alongside the other vendor tools for this protocol. It is an operator capture tool, not a provisioning play, so it does not conflict with the Ansible-only posture — but it **is** superseded by `dhs consumer nmos export` (#837), which cannot produce this class of bug because Go's `json.Marshal` always emits an array for a slice.

## Verification

Against a live registry + node pair:

| | before | after |
|---|---|---|
| `"devices"` starts with | `{` | `[` |
| `"senders"` starts with | `{` | `[` |
| `"self"` starts with | `{` | `{` (unchanged, correct) |
| audit dangling refs | 18,104 (plant) | **0** |

## Definition of done

- [x] tracked under `internal/amwa/assets/`
- [x] verified live: arrays survive, `/self` still an object
- [x] audit reports zero dangling references on the re-captured plant
- [ ] removed once `dhs consumer nmos export` is proven on a real plant

### DIFF
--- a/internal/amwa/assets/nmos-harvest.ps1
+++ b/internal/amwa/assets/nmos-harvest.ps1
@@ -1,0 +1,459 @@
+<#
+  nmos-harvest.ps1 - collect the full x-nmos tree from NMOS devices into assets.
+
+  READ-ONLY: every request is a GET. Nothing is written to any device, no
+  staged/active change, no activation. Safe on production gear.
+
+  Usage:
+    .\nmos-harvest.ps1 -Target 10.44.72.18:3000
+    .\nmos-harvest.ps1 -Target 10.44.72.18:3000,10.44.72.24:3000
+    .\nmos-harvest.ps1 -Target C:\yob\nodes.txt        # one host:port per line
+
+  Output (beside this script unless -Out is given), one folder per device:
+    <host>_<timestamp>\
+       tree.json    - everything in one document (the asset to send back)
+       raw\*.json   - each endpoint verbatim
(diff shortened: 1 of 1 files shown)


by-rune:
Superseded: the harvester path was replaced by the rebuilt stack. Issue #847 stays open for re-triage. Branch kept.

Issue #855, opened 2026-08-25
feat(nmos/node): configurable heartbeat cadence, so the registry's GC is testable


by-rune:
## Why

The node heartbeats every 5 s and there is no way to change it. That is the IS-04 §6.1 default and the right default — but it makes the registry's garbage collector untestable.

Every heartbeat use case needs a cadence:

| case | needs |
|---|---|
| node is evicted when it stops heartbeating | cadence LONGER than the registry's timeout |
| node survives a slow-but-adequate cadence | cadence just under the timeout |
| node re-registers after being GC'd (404 on /health) | eviction, then a heartbeat |
| AMWA test_05 cadence window (5 s ± 0.5 s) | the default, unchanged |
| a lab that wants a fast loop | sub-second |

None of these is reachable against a hard-coded constant. The registry already has `--heartbeat-timeout` and `--gc-interval`; the node has no matching knob.

## What to build

`dhs producer nmos serve --heartbeat DURATION`, defaulting to the spec's 5 s.

The loop currently ticks at 1 s for fast failover detection and throttles actual `/health` POSTs to `HeartbeatInterval - 500ms`. Both need to scale: a 500 ms cadence is swallowed whole by a 500 ms slack, and a sub-2 s cadence needs a faster tick than the beat.

## A gap worth recording

**The threshold is not discoverable.** IS-04 defines the defaults (5 s heartbeat, 12 s timeout) and provides no way for a registry to advertise the value it actually uses:

- DNS-SD TXT carries `api_proto`, `api_ver`, `api_auth`, `pri` — nothing about timing
- `POST /health/nodes/{id}` returns `{"health": "<unix-seconds>"}` — the last-seen time, not the deadline
- the Query API exposes no registry configuration

So a node assumes 5 s and finds out it was wrong only when a heartbeat returns 404. And a **controller** watching the Query API sees a resource disappear with no reason code — a `removed` grain is `pre`-only, identical whether the node deregistered cleanly with DELETE or was garbage-collected. On the customer plant that is exactly the situation with `bm-n-aetac-001`: registered, unreachable, and nothing says why.

Nothing to fix in our code — it is a spec gap — but it belongs in `status.md` §3, and it is the reason the audit's `NMOS-NODE-UNREACHABLE` finding has to be phrased as two possibilities rather than one.

## Definition of done

- [ ] `--heartbeat` on `producer nmos serve`, default 5 s
- [ ] tick rate and throttle slack scale with the configured cadence
[...]

PR #856, opened 2026-08-25
feat(nmos/node): configurable heartbeat cadence, so the registry's GC is testable (#855)

by-rune:
## Why

The node heartbeats every 5 s and there is no way to change it. That is the IS-04 §6.1 default and the right default — but it makes the registry's garbage collector untestable.

Every heartbeat use case needs a cadence:

| case | needs |
|---|---|
| node is evicted when it stops heartbeating | cadence LONGER than the registry's timeout |
| node survives a slow-but-adequate cadence | cadence just under the timeout |
| node re-registers after being GC'd (404 on /health) | eviction, then a heartbeat |
| AMWA test_05 cadence window (5 s ± 0.5 s) | the default, unchanged |
| a lab that wants a fast loop | sub-second |

None of these is reachable against a hard-coded constant. The registry already has `--heartbeat-timeout` and `--gc-interval`; the node has no matching knob.

## What to build

`dhs producer nmos serve --heartbeat DURATION`, defaulting to the spec's 5 s.

The loop currently ticks at 1 s for fast failover detection and throttles actual `/health` POSTs to `HeartbeatInterval - 500ms`. Both need to scale: a 500 ms cadence is swallowed whole by a 500 ms slack, and a sub-2 s cadence needs a faster tick than the beat.

## A gap worth recording

**The threshold is not discoverable.** IS-04 defines the defaults (5 s heartbeat, 12 s timeout) and provides no way for a registry to advertise the value it actually uses:

- DNS-SD TXT carries `api_proto`, `api_ver`, `api_auth`, `pri` — nothing about timing
- `POST /health/nodes/{id}` returns `{"health": "<unix-seconds>"}` — the last-seen time, not the deadline
- the Query API exposes no registry configuration

So a node assumes 5 s and finds out it was wrong only when a heartbeat returns 404. And a **controller** watching the Query API sees a resource disappear with no reason code — a `removed` grain is `pre`-only, identical whether the node deregistered cleanly with DELETE or was garbage-collected. On the customer plant that is exactly the situation with `bm-n-aetac-001`: registered, unreachable, and nothing says why.

Nothing to fix in our code — it is a spec gap — but it belongs in `status.md` §3, and it is the reason the audit's `NMOS-NODE-UNREACHABLE` finding has to be phrased as two possibilities rather than one.

## Definition of done

- [ ] `--heartbeat` on `producer nmos serve`, default 5 s
- [ ] tick rate and throttle slack scale with the configured cadence
- [ ] integration proof: registry `--heartbeat-timeout 6s` + node `--heartbeat 20s` evicts the node; the same pair with `--heartbeat 2s` does not
- [ ] the startup banner prints the cadence in use, not a hard-coded "5s"
- [ ] the non-discoverability of the threshold recorded in `status.md`

### DIFF
--- a/cmd/dhs/cmd_nmos.go
+++ b/cmd/dhs/cmd_nmos.go
@@ -266,7 +266,8 @@ func runNMOSNodeServeLegacy(ctx context.Context, args []string) error {
 	noMDNS := fs.Bool("no-mdns", false, "disable mDNS announce (Mode B / static)")
 	apiVer := fs.String("api-ver", "v1.3", "IS-04 wire version exposed under /x-nmos/node/<v>")
 	priority := fs.Int("priority", 0, "DNS-SD `pri` TXT (0-99 production, 100+ dev)")
-	registry := fs.String("registry", "", "Registration API base URL — when set, the Node POSTs to /resource + heartbeats every 5 s")
+	registry := fs.String("registry", "", "Registration API base URL — when set, the Node POSTs to /resource + heartbeats")
+	heartbeat := fs.Duration("heartbeat", 0, "POST /health cadence (default 5s, IS-04 §6.1). Set it longer than the registry's --heartbeat-timeout to have this node garbage-collected on purpose")
 	if err := fs.Parse(args); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


by-rune:
Closing as superseded-in-part: since #951 the registration client reads the heartbeat cadence live per tick from IS-09 /global (SetHeartbeatIntervalFn), which rewrote the interval plumbing this branch touches - the branch is 8 days stale and CONFLICTING. The remaining asks (--heartbeat CLI flag, ticker/slack scaling for sub-2s cadences, real cadence in the startup banner, registry-GC eviction drill, status.md spec-gap note) are still valid and get rebuilt as a fresh small unit on current main - tracked by #855, scheduled right after the v0.20.0 release cut.

PR #986, opened 2026-09-03
fix(cerebrum-nb): full NB DM walk (self-echo) + watch any node/leaf

### LINKED ISSUE #985: cerebrum-nb: NB DM walk truncated at summary nodes; watch .** exploded — unify on the self-echo rule
## Problem (live-verified on Cerebrum NB, device 2.0.0.2 and 0.0.0.0)

Two NB verbs disagreed on the DM's group-vs-leaf rule, and both were wrong
in opposite directions:

- **`extract`/`tree`** stopped at value-bearing **summary nodes**. On the
  Neuron family `io.sdi` / `io.ip.receivers.audio` answer with a **CSV of
  child UUIDs** (available, STRING) yet have those UUIDs as real children.
  The walk's shortcut (`!available && no value ⇒ node`) classified them as
  leaves → the DM was truncated (`--path io` = 9 objects, vs 360+ when
  seeded one level deeper).
- **`watch --object "X.**"`** used a *separate* probe-walk that recursed
  into every per-stream object and **hard-errored** at the 2000-object cap
  (`expansion exceeded 2000 objects at io.ip.receivers.audio.[uuid].status.legs.[2]`).
  It also subscribed intermediate group handles (which deliver no events)
  and ignored the `--only` filter.

## Fix

- Descent now uses the **NB self-echo rule** (a path that echoes only
  itself is a leaf; otherwise it's a node) via `nbScalarLeaf` +
  `nbMaybeNode`: scalar-typed rows are leaves; typeless/STRING rows that
  could be summary/list nodes are obtained to find out. So `extract`/`tree`
  capture the **full DM with metadata**.
- `watch`'s deep (`.**`) expansion now **reuses that same DM walk**, so it
  registers exactly the leaves (never group handles), applies `--only`, and
  refuses (with "narrow or raise") instead of hard-erroring.

## Verified live (Cerebrum NB 10.6.250.5:40009)

- `tree --device 2.0.0.2 --sub-device 0 --path io.madi` → 48 objects, full
  per-UUID depth + metadata (type/enum/access/value).
- `extract … io.madi` → DM of 48 objects with metadata.
- `watch --device 0.0.0.0 --sub-device 0 --object "ComputerOverview.**" --initial`
  → subscribes 18 leaves, prints baseline then live changes (human-readable).

by-rune:
Closes #985.

## What

The Cerebrum **Northbound** DM walk now recurses the full model with
metadata, and `watch` can target any node or leaf without exploding —
both grounded in live NB behaviour, not assumptions.

## The bug (two verbs, opposite failures)

The NB tree's group-vs-leaf rule is **self-echo**: obtain a path — one row
echoing itself ⇒ leaf; other rows ⇒ node. Two verbs each broke it:

- `extract`/`tree` used a shortcut (`!available && no value ⇒ node`), so
  **value-bearing summary nodes** were mis-read as leaves. On the Neuron
  family `io.sdi` / `io.ip.receivers.audio` answer with a CSV of child
  UUIDs (available, STRING) **and** have those UUIDs as children →
  `--path io` returned 9 objects; seeding one level deeper returned 360+.
  The DM was truncated.
- `watch --object "X.**"` used a *separate* probe-walk that recursed into
  every per-stream object, **hard-errored** at the 2000-object cap, also
  subscribed event-less group handles and ignored `--only`.

## Fix

- `nbScalarLeaf` + `nbMaybeNode` classify by the self-echo rule: scalar
  descriptors (INTEGER/ENUM/BOOL/FLOAT/…) are leaves; typeless handles and
  STRING rows that could be summary/list nodes (empty, comma-list, bare
  UUID) are obtained to find out. `extract`/`tree` capture the full DM +
  metadata.
- `watch`'s deep expansion reuses that same walk: leaves only, `--only`
  applied, and a subtree too large to enumerate is refused (narrow/raise)
  rather than erroring mid-dive.
- Removed the now-dead `cerebrumExpand` probe-walk, `hasOtherPaths`,
  `maxExpandDepth`; swapped the obsolete `hasOtherPaths` test for
  `TestNBNodeClassification` over the new classifiers.

## Verified live (Cerebrum NB 10.6.250.5:40009)

- `tree --device 2.0.0.2 --sub-device 0 --path io.madi` → 48 objects, full
  per-UUID depth + metadata.
- `extract … io.madi` → DM of 48 objects with metadata (access/type/value).
- `tree --device 0.0.0.0 --sub-device 0 --path ComputerOverview` → full
  Memory/Network subtree with types.
- `watch --device 0.0.0.0 --sub-device 0 --object "ComputerOverview.**" --initial`
  → subscribes 18 leaves, prints baseline then live changes, human-readable.

## Follow-up (not in scope)

Under a large initial burst the dispatcher logs `dropped reply (channel
full)` — a reply-channel buffer size issue this fix merely exposes (watch
now actually subscribes many leaves). Tracked separately.

### DIFF
--- a/cmd/dhs/cmd_cerebrum.go
+++ b/cmd/dhs/cmd_cerebrum.go
@@ -1899,14 +1899,9 @@ func cerebrumChildren(sess *cerebrum.Session, timeout time.Duration, device stri
 	return got.Device.ObjectValues, nil
 }
 
-// maxExpandDepth bounds "**". The deepest live NB tree seen is a node's
-// Interfaces/Devices at two levels; the cap is a guard against a cyclic
-// or pathological tree, not a real limit.
-const maxExpandDepth = 8
-
-// maxExpandObjects bounds what one expansion may subscribe.
+// maxExpandObjects bounds what one "**" expansion may subscribe.
 //
-// A deep walk probes every child, so "Nodes.**" on a live NOC is 68
+// A deep walk descends every node, so "Nodes.**" on a live NOC is 68
 // nodes x ~16 fields, then into Interfaces and Devices, then into each
 // device's Senders/Receivers/Sources — tens of thousands of obtains
 // against a production control system, from one careless command.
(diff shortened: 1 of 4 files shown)


by-rune:
Closing as **already merged** � not abandoned.

Rebasing this branch onto current `main` produces a tree **byte-identical to
main**: zero commits remaining, zero files changed. All three commits are
already in, carried by the #989 squash:

- `ef377e9d` recurse the full NB DM (self-echo rule)
- `d8160909` extract falls back on identity
- `6bedb492` carry the descriptor onto watch change events

`cerebrumExpand` on main is this PR's own *later* form � the
`[]codec.DeviceObjectValue` descriptor-carrying signature from `6bedb492`,
not the earlier `[]string` one. The branch showed CONFLICTING only because it
was cut from old main and never rebased; the conflict was this work against a
newer version of itself.

Verified:

    git rebase --onto origin/main ... fix/cerebrum-nb-selfecho-dm-walk
    git rev-parse HEAD^{tree} == git rev-parse origin/main^{tree}   # identical

Nothing is lost by closing. Issue #985 is delivered.

PR #1006, opened 2026-09-06
refactor(probel-sw02p): the first connector that cannot open a socket

by-rune:
Step 3 of the sweep — **one connector**, proving the pattern end to end before it's repeated nine more times.

| Layer | Change |
|---|---|
| consumer | `Plugin` holds `deps.Net` and `deps.Metrics`; `session.Dial` takes the Net instead of building a `TCPDialer` |
| session | `Dial(ctx, n transport.Net, addr, logger, cfg)` — this package no longer decides *how* a socket is made, only what to say over it |
| provider | `newServer(deps, tree)`; `Serve` binds through `s.net.Listen` |

After this the package contains **no `net.Dial`, no `net.Listen`, no `ListenConfig`, no `transport.TCPDialer`**. The only remaining `net.Listener` mentions are the stdlib *type* in `ServeListener`'s signature — the seam an injected listener arrives through, not a socket being made.

`metrics.NewConnector()` is gone from both halves: the counter set is supplied, so the process can scrape every connector from one place instead of each owning a registry nobody else can reach.

## Two things worth stating rather than hiding

**A regression I introduced and caught.** `Connect` used to construct its own metrics, so a `Plugin` built directly — which several tests do — had a working `Metrics()`. Reading `deps.Metrics` unconditionally made that nil, and three consumer tests failed on it. Fixed with the same shape `WithDefaults` uses: a hand-built `Plugin` still gets a counter set. The factory path always injects one; the guard is for callers that predate `Deps`.

**`session.Dial` keeps a nil-Net fallback** that builds a default transport. That's a convenience for direct callers, not a hole in the guarantee: the plugin always injects, and the architecture test still sees no socket construction in the protocol package. Removing it would only push `transport.New` into every test.

## Verification

| | |
|---|---|
| codec / session / consumer / provider | **100% each** |
| build + vet + full suite | green |
| coverage floors | **32/32** |
| `golangci-lint` | 0 issues |

One process note: my first attempt at rewriting the provider test call sites used a regex whose `[^,]+` matched across newlines and mangled six files. Reverted and redone line-scoped — worth mentioning because the revert also undid the `server.go` change, which I had to re-apply.

**Nine connectors follow the same three steps**, and each deletes rather than moves.

### DIFF
--- a/internal/probel-sw02p/consumer/plugin.go
+++ b/internal/probel-sw02p/consumer/plugin.go
@@ -68,11 +68,10 @@ func (f *Factory) Meta() consumer.ProtocolMeta {
 	}
 }
 
-// New constructs a fresh consumer plugin bound to the given logger.
+// New constructs a fresh consumer plugin bound to the injected dependencies.
 func (f *Factory) New(deps plugin.Deps) consumer.Protocol {
 	deps = deps.WithDefaults()
-	logger := deps.Logger
-	return &Plugin{logger: logger}
+	return &Plugin{logger: deps.Logger, net: deps.Net, metrics: deps.Metrics}
 }
 
 // MatrixConfig holds the externally-supplied matrix shape. SW-P-02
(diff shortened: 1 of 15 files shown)


by-rune:
Closing without merging.

Not because the change is wrong — it is green and the pattern is sound — but because it migrated 1 connector of 10, and a tree with two connector shapes in it is worse than a tree with one consistent shape, even a less ideal one.

The transport refactor is parked. The Deps and Net contracts stay on main (merged in #1003/#1004) and cost nothing while unused; if the sweep is ever resumed, the work in this branch is recoverable from this PR.

Priority returns to real-device integration testing.

Issue #1009, opened 2026-09-07
feat(snell-rollcall): 16/32-bit consumer + provider connector (epic)


by-rune:
## Goal

Build the **Snell RollCall** connector in Go: **consumer AND provider**, both wire
generations (16-bit and 32-bit / `SV_LONGSTR`), stdlib-only codec, to the ADR-0025
six-deliverable bar.

Validated groundwork already exists — see
`internal/snell-rollcall/docs/audit-2026-09-07.md`:

- live oracle on `10.6.250.105` (RollProxy `:2050`, Centra `:2057`, 15 real units)
- consumer path proven read/write/subscribe against the real stack
- a throwaway producer spike was **rendered and driven by RollCall Control Panel 4.12.48**
- 560 menu lines + 1 623 frames captured as fixtures

## Scope

- `internal/snell-rollcall/{codec,session,consumer,provider,wireshark,integration,testdata,docs}`
- Both generations on one connector; per-session negotiation from `rService`
- TCP only (IPShare/IP-Proxy). **Out of scope:** RollNet/ArcNet, serial, I2C

## Units (one PR each, ADR-0013/0025)

- [x] 1. Audit gate + `CLAUDE.md` + docs scaffold (no code) — #1011
- [x] 2. codec: framer, `FullAddress`, packet-type table, 16-bit payloads — #1014
- [x] 3. codec: 32-bit payloads (65-71), long strings, DTP, router arithmetic, unit-id table — #1017
- [x] 4. session: link, sessions, back channel (one-active-message + ACK), keepalive — #1019
- [x] 4b. codec: file, time, log, stream, display, group services + spec coverage matrix — #1021
- [x] 5. **Deliverable 1** consumer — #1023
- [ ] 6. **Deliverable 2** provider
- [ ] 7. **Deliverable 5** Wireshark dissector (superset of vendor `rollcall.dll`)
- [ ] 8. **Deliverable 3** integration: Go `-tags integration` + **Ansible plays only**
- [ ] 9. **Deliverables 4+6** DM/manifest fixture + replay set
- [ ] 10. Docs set + README coverage table

## Binding constraints

- ADR-0006: `codec/` stdlib-only, lift-ready, no `dhs/*` imports
- ADR-0025: integration driven by **Ansible only — no `.sh`, no `.ps1`, no other language**
- ADR-0025: producer must additionally be driven by a **third-party client**
[...]

PR #1014, opened 2026-09-07
feat(snell-rollcall): codec part 1 — framer, addressing, 16-bit payloads

### LINKED ISSUE #1012: feat(snell-rollcall): codec part 1 - framer, addressing, packet-type table, 16-bit payloads
Part of #1009.

**Unit 2 of 10.** First half of the stdlib-only codec (ADR-0006: no `dhs/*`
imports, lift-to-own-repo ready).

## Scope

- `frame.go` — transmission header + `MESSAGE_STR` + `ROLLHEADER_STR`, and an
  `io.Reader` framer with **1-byte resync** (the C library skips 4, which can
  never re-align on an odd offset)
- `addr.go` — `FullAddress`, `NNNN-UU-PP:SS` parse/format, bridge-nibble
  validation, the reserved indices (`0` blind, `0xFE` logging, **`0xFF` unknown**)
- `pkttype.go` — all 72 packet types, names, and the per-type property table
  (broadcastable / INVSESS-able / valid-blind-reply) taken from the vendor
  `MsgHandlers.c` table
- 16-bit payload structs: `CONNECT_STR`, `TERMSESS_STR`, `CLEARSESS_STR`,
  `ID_STR`, `VERSION_STR`, `STATUS_STR`, `DEVICEINFO_STR`, `GETFSTAT_STR`,
  `FUNCSTATUS_STR` (+ flag-driven tail), `FUNC_STR`, `FUNCSTYLE_STR`,
  `BLOCKHEADER_STR`, `GETNEXT_STR`, `DISP_STR`, `WAIT_STR`, `TIME_STR`
- `errors.go`, `strings.go` (fixed-field vs NUL-terminated rules)

## Not in scope

32-bit types 65-71, Data Transfer Params, router arithmetic, the unit-id table
(unit 3); session layer (unit 4); anything above (units 5+).

## Test sources

1. **Spec bytes** — expected-byte tables from Rev 14 §11, never from working code
2. **Real frames** — `docs/captures/rollcall-live-2026-09-07.hex`, 1 623 frames
   captured from the live oracle, decoded and round-tripped

## Acceptance

- `go test -race` green; **100% statement coverage** on the package, locked as a
  CI floor
- round-trip: every captured frame decodes and re-encodes byte-identically
- no import outside stdlib

by-rune:
Closes #1012. Part of #1009 (unit 2 of 10).

Stacked on #1011 (the audit and wire context), so the base is that branch
rather than `main`. It retargets to `main` automatically when #1011 merges.

## Ported or reimplemented?

Reimplemented. This is a clean-room Go package written from the RollCall
Technical Specification Revision 14, with the vendor C library used as a
reference for the behaviours the specification leaves open. No vendor source
was translated, no vendor object is linked, and there is no cgo. Every struct
carries the specification section it comes from, and where a rule comes from
the vendor library instead the comment names the file it was read from.

## What is here

| Area | Contents |
| --- | --- |
| Framing | 4-byte transmission header, 14-byte message header, 2-byte roll header, encoder, and a resynchronising stream reader |
| Addressing | `NNNN-UU-PP:SS` parse and print, bridge routing, hop counting, route validation |
| Catalogue | all 72 packet types with their broadcast, blind-reply and generation properties |
| Bit fields | service mask, user level, unit status, value mode, menu style, termination code |
| Strings | fixed 20-byte fields and NUL-terminated long strings, with the UTF-8 truncation rule |
| Payloads | identity, session, control and menu structures of the 16-bit generation |

The 32-bit generation, the file and display services, and the router and
unit-id tables arrive in unit 3.

## Wire rules this pins

Four points where the specification is silent, ambiguous, or contradicted by
every shipping device. Each has a named test.

**The unconnected session index is 255, not -1.** The field is signed, so the
obvious encoding of "unknown" puts `0xFFFF` on the wire. A lenient proxy
answers it anyway; a real device faults. This crashed a vendor simulator three
times during the audit before the cause was found.

**Framing resynchronises one byte at a time.** The vendor library discards four
bytes per attempt, which cannot re-align a stream that slipped by an odd
number. Resyncs are counted so the session layer can raise a compliance event.

**Bytes after a NUL in a fixed field are undefined.** Live Centra units and the
vendor test client both leave `0xCD` there. That is compliant, so it is never
reported and never compared. Fixtures that depend on it would be
device-specific.

**A level-gated menu line is substituted, not removed.** The server replaces it
with a hidden, disabled Data line carrying command 0 and the text "Reserved".
Line counts are identical at every user level, so comparing counts to detect
gating always reports no difference. `AccessGated` matches the whole pattern.

## Defects found while testing

**The reader could not assemble a spec-legal large frame.** Its buffer held 880
bytes while `DecodeFrame` accepts up to 1574, the specification's limit. A peer
sending a legal 1570-byte message would have hit a buffer-full error and lost
the connection. The buffer is now sized from `MaxDecodeFrame`, and a test feeds
a maximum-size frame through in 200-byte pieces.

**`Address.String` hid the defect it should have exposed.** It masked the
session index to 8 bits, so the malformed `0xFFFF` printed as a plausible
`0FF`. It now prints the full 16-bit field.

Both are fixed in this PR.

## Verification

```
ok  dhs/internal/snell-rollcall/codec   coverage: 100.0% of statements
```

- 109 tests, table-driven, expected bytes taken from the specification.
- **1620 frames** captured from Snell equipment on 10.6.250.105 decode and
  re-encode byte-identically across **39 packet types**. These vectors were
  produced by the vendor's own stack, not by this implementation, which is
  what makes them worth something.
- 784 device-originated frames are asserted to carry a well-formed session
  index. The 21 probe frames still carrying the pre-fix encoding are asserted
  to be ours, so the evidence stays visible instead of being deleted.
- 100% statement coverage, added to the per-package floor in CI.
- stdlib-only, no `dhs/*` import (ADR-0006). Builds for linux, darwin,
  windows and freebsd on amd64 and arm64 with cgo disabled (ADR-0016).

Reaching 100% meant removing error paths that could never run rather than
writing tests that could never exercise them: where an outer length check
already guarantees a structure's size, the nested decodes are now infallible
helpers. That made the decoders smaller as well as fully covered.

## Not in this PR

Consumer, provider, session layer, dissector and integration plays are units
4 onward. The integration tier will be driven by Ansible only, per the
standing instruction.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,7 @@ jobs:
       # "^dhs/<pkg>/<file>.go:" so nested packages stay separate; a floor
       # package with no rows at all is a hard failure (renamed package
       # guard). On a miss the exact uncovered blocks are printed.
-      - name: Coverage floor (acp1 + acp2 + emberplus)
+      - name: Coverage floor (per-package)
         if: matrix.os == 'ubuntu-latest'
         run: |
           set -e
@@ -168,6 +168,7 @@ jobs:
           check internal/osc/consumer 100
           check internal/osc/provider 100
           check internal/cerebrum-nb/codec 100
+          check internal/snell-rollcall/codec 100
           check internal/transport/ws 100
           check internal/transport/http 100
           check internal/auth 100
--- a/internal/snell-rollcall/codec/addr.go
+++ b/internal/snell-rollcall/codec/addr.go
@@ -1,0 +1,239 @@
+package codec
+
+import (
+	"encoding/binary"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// AddrSize is the wire size of a full address (spec 11.1.1).
+const AddrSize = 6
+
+// Reserved session indices (spec 5.4, 12.3).
+const (
+	// IndexBlind addresses a unit without a session. The server treats such
+	// requests as UL_SUPERVISOR (spec 9.17).
+	IndexBlind int16 = 0
+
+	// IndexLogging is reserved for the logging service (vendor
+	// CoreServiceAPI.h LOGGING_SESSION).
+	IndexLogging int16 = 0xFE
+
+	// IndexUnknown is UNKNOWNSESS: unconnected traffic, broadcasts, and the
+	// destination of every Call.
+	//
+	// It is the value 255. Writing "unknown" as -1 is the single easiest
+	// mistake to make here, because the field is signed; -1 encodes as
+	// 0xFFFF, which a lenient proxy answers anyway and a real device faults
+	// on. See internal/snell-rollcall/CLAUDE.md "What NOT to do" #1.
+	IndexUnknown int16 = 0xFF
+)
+
+// Unit address ranges (spec 5.1).
+const (
+	UnitBroadcast   uint8 = 0x00 // also "the unit at the other end of this link"
+	UnitBridgeFirst uint8 = 0x01
+	UnitBridgeLast  uint8 = 0x0F
+	UnitFirst       uint8 = 0x10
+)
+
+// Address is a RollCall address: a device (Net, Unit, Port) plus a session
+// index. Written NNNN-UU-PP:SS.
+//
+// Spec 11.1.1 FULLADDRESS_STR, 6 bytes:
+//
+//	UINT16 rNet    4 nibbles of network route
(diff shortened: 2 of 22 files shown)


by-rune:

PR #1017, opened 2026-09-07
feat(snell-rollcall): codec part 2 — 32-bit payloads, DTP, router arithmetic, unit ids

### LINKED ISSUE #1016: feat(snell-rollcall): codec part 2 — 32-bit payloads, DTP, router arithmetic, unit-id table
Unit 3 of #1009. Follows #1012 (codec part 1).

## Scope

Everything the 32-bit generation needs, plus the two data models that only
exist there.

### 32-bit payloads (packet types 65-71)

From `Long String and 32-bit Command Number support.docx` (V1, 2 June 2014),
with field widths confirmed against `rc3comm.h` and `RC3TYPES.H`. Structures
are `#pragma pack(2)`, so there is no padding between fields.

| Struct | Size | Carried by |
| --- | --- | --- |
| `MENUREQ_STR` | 4 | `SP_GETMENUCOUNT` (65), `SP_GETMENUITEM` (67) |
| `MENUSIZE_STR` | 8 | `SP_RETMENUCOUNT` (66) |
| `MENUITEM_STR` | 22 + two NUL-terminated strings | `SP_RETMENUITEM` (68) |
| `GETVALUE_STR` | 4 | `SP_GETVALUE` (69) |
| `VALUE_STR` | 12 + optional string or data | `SP_SETVALUE` (70), `SP_RETVALUE` (71) |

Differences from the 16-bit generation that the code must honour:

- `SP_GETMENUITEM` takes an **absolute** menu index. `SP_GETNEXTPKT` takes an
  offset from the base of a previous `SP_GETFUNC`, and relies on the server
  holding state. The new one is stateless, so a client may fetch items in any
  order and in parallel.
- `VALUE_STR` has a **declared** `rMatchID` field, at a fixed offset before the
  mode. In `FUNCSTATUS_STR` the match ID trails the payload and only exists
  when the flag is set, which is what forces the fixed-width string there.
- There is no 32-bit equivalent of `SP_FUNCSTYLECHG`. A back-channel menu
  update is always a full `SP_RETMENUITEM`.
- `Range_t` is `INT32`, so ranges and values are signed even though the 2014
  paper writes `UINT32`.

### Data Transfer Params

The value encoding the router command set uses. A count byte whose bit 7 means
string-safe, then items typed 1 uint-array, 2 bitmap, 3 false, 4 true, 5 uint,
6 string. Integers are variable-length, 7 bits per byte, least significant
first, continuation in bit 7. String-safe blocks escape `00` as `FF FD` and
`FF` as `FF FE`.

Lands as `codec/dtp`, a self-contained package.

### Router arithmetic

The Full Control Command Set places a matrix in a flat 32-bit command space:

```
MatrixBase(m) = CMD_MATRIX_BASE + (m-1) * CMD_MATRIX_STEP
```

with the same shape again per level and per destination. Also the source pin
packing (`matrix<<24 | level<<16 | src`), the protect word (bit 0 protect,
bits 8..23 protect id, bit 24 master), the names-file layouts (8- and 32-char
fixed arrays, 32-bit little-endian association mappings), and the CRC32 with
polynomial 0x04C11DB7 used to validate a cached names file.

Lands as `codec/router`, arithmetic only. No I/O.

### Unit-id table

About 700 unit-type ids mapping to product names, generated from `rc3id.h`
and `RollCall_IDs.xml`. Feeds identity in the manifest and DM per ADR-0022,
and the dissector Info column.

## Acceptance

- `go test -race` green.
- **100% statement coverage** on every package added, locked as a CI floor.
- Expected bytes in tests come from the 2014 paper and the router document,
  never from this implementation.
- Round-trip: every 32-bit frame in the captured trace decodes and re-encodes
  byte-identically.
- No import outside the standard library, and no `dhs/*` import (ADR-0006).
- The generator for the unit-id table is committed with its output, so the
  table can be regenerated rather than hand-edited.

## Out of scope

Session layer, consumer, provider. The router package is arithmetic and
layout only; fetching a names file over `SV_FILE` belongs to the consumer.

by-rune:
Closes #1016. Part of #1009 (unit 3 of 10).

Stacked on #1014, which is stacked on #1011. Each retargets automatically as
the one below it merges.

## Ported or reimplemented?

Reimplemented, same as unit 2. Written from the 2014 long-string paper and the
Full Control Command Set document, with field widths confirmed against
`rc3comm.h` and `RC3TYPES.H`. No vendor code was translated and nothing is
linked.

## What is here

| Package | Contents |
| --- | --- |
| `codec` | packet types 65-71 and their five structures, plus the 770-entry unit-type table |
| `codec/dtp` | Data Transfer Params: the parameter block the router command set carries values in |
| `codec/router` | the flat 32-bit command space, the source pin and protect word, and the layouts of the files names come from |

## Worked examples reproduced byte for byte

Three rules that the documents state only through a worked example. Each
example is now a test, and each passes exactly as printed.

**Data Transfer Params.** Both examples from the router document, including
the six-item block that exercises every item type in one go. Integers are
seven bits per byte, least significant first, continuation in bit 7.

**The names-file checksum.** This is the one most worth having pinned. It is
not a hash of the file. Each name is hashed padded to its field width and
followed by its index as four little-endian bytes, and those hashes are added
with 32-bit wrapping. The document gives four per-name values and their total:

| Name | Expected | Got |
| --- | --- | --- |
| m01a01 | `a2b261ef` | `a2b261ef` |
| m01a02 | `2be61c17` | `2be61c17` |
| m01d01 | `422cec2b` | `422cec2b` |
| m01d02 | `cb7891d3` | `cb7891d3` |
| sum | `dc3dfc04` | `dc3dfc04` |

Which also settles what the document leaves vague: the hash is the ordinary
IEEE CRC-32, so `hash/crc32` computes it and no custom implementation is
needed. Summing rather than chaining is deliberate, and a test shows why: a
controller can swap one name's contribution when that name changes instead of
re-reading the file.

**The string-safe escape.** Its exemption for "the first or last byte" is
about the block, not the body. The block's first byte is the count, which
carries the string-safe marker and is never escaped; so within the body only
the final byte is exempt. Getting this backwards produces a block a vendor
decoder reads as one item short.

## Two structural differences a decoder must not paper over

**The match id moved.** `VALUE_STR` declares it at a fixed offset before the
mode, so it is always on the wire and the string after it stays
variable-length in every combination. In `FUNCSTATUS_STR` the id trails the
payload and exists only when its flag is set, which is precisely why the
string before it has to become fixed-width. Same flag, opposite consequence.

**The menu request became stateless.** `SP_GETMENUITEM` takes an absolute
index. `SP_GETNEXTPKT` takes an offset from the base of a preceding
`SP_GETFUNC` and relies on the server holding that base, so items must be
walked in order on one session. The new form can be fetched in any order and
in parallel, which is what makes a large menu bearable.

## Projecting between generations

A provider holds one tree and answers both kinds of client, so the two
directions are here. Widening is lossless. Narrowing truncates a label,
because a shortened name still names the line, and refuses a number, because
a truncated command number addresses a different command on the device. The
router command set is exactly why that distinction matters: its commands sit
well above `0xFFFF` by construction.

## Verification

```
ok  dhs/internal/snell-rollcall/codec          coverage: 100.0% of statements
ok  dhs/internal/snell-rollcall/codec/dtp      coverage: 100.0% of statements
ok  dhs/internal/snell-rollcall/codec/router   coverage: 100.0% of statements
```

From the live trace on 10.6.250.105, frames sent by Snell equipment:

| Structure | Frames | Result |
| --- | ---: | --- |
| RETMENUITEM | 562 | re-encode byte-identically |
| RETVALUE / SETVALUE | 65 | re-encode byte-identically |

Round-tripping rather than merely decoding is the point. A field offset that
is wrong by two still decodes without error, it just reads the wrong fields;
only re-encoding to the same bytes rules that out. Of the 562 menu items, 557
carry text and 49 are containers, which is the shape a real menu has and a
misread structure would not produce.

All three packages are stdlib-only with no `dhs/*` import, and build for five
OS/arch pairs with cgo disabled.

## The unit-id table

Generated by `tools/gen-rollcall-ids` from the vendor's `RollCall_IDs.xml`,
which is the file the vendor's own `genfiles.py` reads to produce `rc3id.h`.
The generator is committed with its output so the table is regenerated rather
than hand-edited.

Three ids appear in both sections of that database naming different products,
which is a genuine collision in vendor data rather than a parsing mistake:

| id | RollCall section | Vistek section |
| --- | --- | --- |
| 78 | ID_MDAVE_1 | (unnamed) |
| 606 | Sirius 830 chassis | V606 Rack Controller Module |
| 608 | Sirius 850 chassis | V608 Rack Controller Module |

We take the RollCall section, as the vendor generator does, and a test pins
that choice so it cannot drift silently.

The table is a sorted array searched by bisection, not a map literal. At 770
entries a map costs start-up time and heap for no benefit; the array is static
data with neither.

## Not validated

The offsets inside the per-matrix, per-level, per-source and per-destination
tables are read from the order the router document lists their commands in.
It gives no numbers for them, only for the root table, and says the rest are
"defined in a similar fashion". The audit oracle answers NACK to command 100,
so it exposes no routing interface to check them against.

This is stated in the package comment rather than left implicit. Confirming it
needs a real router controller, or the simulator reconfigured with
`CENTRA_SIMULATED_ROUTER`, and belongs to the integration unit.


by-rune:

PR #1019, opened 2026-09-07
feat(snell-rollcall): session layer — link, sessions, back channel, keepalive

### LINKED ISSUE #1018: feat(snell-rollcall): session layer — link, sessions, back channel, keepalive
Unit 4 of #1009. Follows #1017.

The protocol state machine, transport-agnostic. It takes a `net.Conn` and the
standard dependency set, and it is the same machine on both sides with the
roles flipped, which is why it sits below `consumer/` and `provider/` rather
than inside either.

## Scope

`internal/snell-rollcall/session/`

| File | Contents |
| --- | --- |
| `link.go` | one TCP link: read and write loops, frame dispatch, IPShare address rewriting, reconnect |
| `session.go` | CALL, ACK, TERM; the one-active-message queue; the 3 s timeout and its WAIT extension; five-strike death |
| `backchannel.go` | push receipt and acknowledgement on the consumer side, the push queue on the provider side |
| `identity.go` | our own DeviceInfo and the IAM schedule |
| `listsvc.go` | the BLOCKHEADER and GETNEXTPKT multi-packet walker |
| `keepalive.go` | the idle probe |

## Rules it has to hold

**One active message per channel per session.** A request occupies the slot
until its reply arrives, and nothing else may be sent on that channel
meanwhile. The keepalive probe is an active message like any other and goes
through the same queue; injecting it beside a pending request matches it
head-of-queue against the wrong reply.

**Three seconds, extended by WAIT.** The vendor library never implements the
extension and answers NACK to the message; we implement it, because a server
that asks for more time and is then abandoned leaves a session behind.

**Five consecutive failures ends the session.** The vendor drops it without
sending TERM, which is how units run out of sessions.

**Always TERM.** On exit, on error, on context cancellation. Servers do not
time idle sessions out, so a leaked session is leaked until the unit reboots.

**The client session index is sticky.** The index a server returns in its ACK
is the one every later message on that session must carry. Getting this wrong
produced SP_INVSESS and silently lost every push during the audit.

**IPShare address rewriting.** A client sends a zeroed source address and the
gateway stamps its own port number in; the client restores its address on the
way back. A peer that does not do this is rejected by real gateways and by the
Control Panel. The first message after connect is GETDEVINFO to the
unconnected broadcast address, whose reply reveals the gateway address.

**The back channel is off until asked for.** BKCHNREADY enables it, and
control pushes additionally need REPFCHG. On enable the server flushes every
changed value, so a client must be ready for a burst.

## Keepalive

Measured during the audit rather than inferred: KEEPALIVE is answered with ACK
by both a real device and the proxy, on both an unconnected and a connected
session, in 4 ms. Zero unsolicited traffic arrives on an idle connection, so
silence carries no information and liveness must be probed.

Defaults: probe every 15 s when idle, 3 s per probe, three consecutive misses
drops and reconnects with backoff. GETDEVINFO stays as the fallback probe for
a peer that ignores type 28.

## Acceptance

- `go test -race` green.
- **100% statement coverage**, locked as a CI floor.
- No sleeping in tests: every timeout, probe and backoff is driven through the
  injected clock.
- Takes `plugin.Deps`, so it cannot open a socket or read the wall clock by
  itself.
- Never imports `consumer/` or `provider/`, and neither imports the other.

## Out of scope

The consumer and provider themselves, and the verbs they expose. This unit
ends at a session you can send a request on and receive a push from.

by-rune:
Closes #1018. Part of #1009 (unit 4 of 10).

Stacked on #1017. It retargets automatically as the one below it merges.

## What this is

The protocol state machine. It sits below `consumer/` and `provider/` because
it is the same machine on both sides with the roles exchanged: a link carrying
frames, sessions multiplexed over it by index, one message in flight per
channel, and a back channel whose pushes are themselves requests. The vendor's
own `Core/` is organised the same way for the same reason.

It is handed a `net.Conn` and never opens one. Every test runs over `net.Pipe`
with no socket, no port and no listener, and time comes from the injected
clock, so a three-second timeout is tested by advancing a fake rather than by
waiting three seconds. There is no `time.Sleep` in the package's logic, and
none in its tests except the poll that waits for a timer to be armed.

## The rule everything else follows

One message in flight per channel per session. A peer matches replies
head-of-queue, so two requests on the wire at once are not merely reordered,
they are answered to the wrong caller. Acquiring the slot is the only way to
send, which makes the rule structural rather than a convention every call site
has to remember.

The keepalive probe obeys it too. Injecting a probe beside a pending request
is not a shortcut, it is the same bug.

## Five behaviours the vendor gets wrong or leaves out

**Wait extends the deadline.** The vendor library answers Nack to it and never
implements the extension, so a server that asks for more time is abandoned
mid-operation and its session is left behind.

**Term is always sent** — on close, on error, on cancellation. And directly,
not through a queue slot that a request which will never be answered may still
be holding: waiting for that slot is exactly how a session ends up abandoned
instead of closed. Servers do not time idle sessions out, which the vendor's
own source comments warn about, so a session we walk away from is leaked until
the unit reboots.

**Five consecutive failures end a session and say so.** The vendor drops it
without sending Term, which is how units run out of sessions.

**A push is acknowledged by the application, not on its behalf.** The
acknowledgement is what asks for the next push, so sending it before the
application has taken this one throws away the only flow control the protocol
offers. A full queue withholds acknowledgements and stalls the peer, which is
what the protocol intends.

**The two session indices are not interchangeable.** Ours goes in the source
of what we send and the destination of what is pushed to us. The peer's, from
its acknowledgement, goes in the destination of what we send. Swapping them
produced SP_INVSESS during the audit and silently lost every back-channel push.
There is a test for each of the three ways it went wrong.

## Address rewriting

A TCP client knows neither address. Its own is assigned by the gateway, which
stamps a port drawn from its connection slots over the zeroed source we send;
the gateway's own is wherever the reply comes from. `Handshake` learns both,
and a peer that skips this is rejected by real gateways and by the Control
Panel.

Both are read from the frame header, never from the DeviceInfo payload: the
address inside one is not rewritten as a message crosses a bridge, so its
route is meaningless. There is a test with a payload and a header that
disagree.

## Keepalive, measured rather than inferred

From the audit against the live stack:

| Probe | Direct to the device | Via the proxy |
| --- | --- | --- |
| KeepAlive on the unconnected index | Ack, 4 ms | Ack, 4 ms |
| KeepAlive on a connected session | Ack, 4 ms | Ack, 4 ms |
| Unsolicited frames during 90 s idle | 0 | 0 |
| Session alive after 90 s of silence | yes | yes |

Zero unsolicited traffic means silence carries no information, so liveness has
to be probed. The probe goes on the session when there is one, because that
proves the session still exists and the session is what actually breaks.
Defaults: every 15 s, 3 s per probe, three misses ends the link.

## Two defects fixed while building it

**The reply deadline leaked a timer per Wait.** It used `After` in a retry
loop, so every extension left an armed timer behind until it fired. It is now
a stoppable timer. That also let the tests assert an exact armed-timer count
rather than a minimum, which is what makes them deterministic.

**Allocating an index and registering the session were two steps.** Between
them was a window in which the peer's acknowledgement, which is addressed to
that index, could arrive with nothing listening on it. They are now one
operation under one lock.

## Verification

```
ok  dhs/internal/snell-rollcall/session   coverage: 100.0% of statements
```

- 87 tests, 100% statement coverage, added to the CI floor.
- Ten consecutive runs, no flake.
- Imports `context`, `net`, `sync`, `time` and four dhs packages: codec, clock,
  metrics and the dependency set. It cannot reach `consumer/` or `provider/`,
  and they cannot reach each other.
- Builds for five OS/arch pairs with cgo disabled.

## Also in this PR

Two codec helpers are exported because the session layer needs them:
`CString`, for the optional text a Nack or a Wait may carry, and
`TruncateFixed`, for naming ourselves in an identity a peer will display. A
client that cannot connect because its own name is long is a worse outcome
than one that appears under a shortened name.

## Not in this PR

The verbs. This package hands you a session you can send a request on and
receive a push from; deciding which requests to send is the consumer's job,
and answering them is the provider's. Those are units 5 and 6.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,7 @@ jobs:
           check internal/snell-rollcall/codec 100
           check internal/snell-rollcall/codec/dtp 100
           check internal/snell-rollcall/codec/router 100
+          check internal/snell-rollcall/session 100
           check internal/transport/ws 100
           check internal/transport/http 100
           check internal/auth 100
--- a/internal/snell-rollcall/codec/payload_control.go
+++ b/internal/snell-rollcall/codec/payload_control.go
@@ -139,7 +139,7 @@ func DecodeFuncStatus(b []byte) (FuncStatus, error) {
 			f.Text = fixedString(b[off : off+MaxTextSize])
 			off += MaxTextSize
 		} else {
-			s, n := cString(b[off:])
+			s, n := CString(b[off:])
 			f.Text = s
 			off += n
 		}
--- a/internal/snell-rollcall/codec/payload_control32.go
+++ b/internal/snell-rollcall/codec/payload_control32.go
@@ -109,7 +109,7 @@ func DecodeValue(b []byte) (Value, error) {
 
 	off := ValueSize
 	if v.Mode.Has(ModeString) {
-		s, n := cString(b[off:])
+		s, n := CString(b[off:])
 		v.Text = s
 		off += n
 	}
@@ -142,7 +142,7 @@ func (v Value) ToFuncStatus() (FuncStatus, error) {
 		Command: uint16(v.Command),
 		Mode:    v.Mode,
 		Value:   v.Val,
-		Text:    truncateFixed(v.Text, MaxTextSize),
+		Text:    TruncateFixed(v.Text, MaxTextSize),
 		Data:    v.Data,
 		MatchID: v.MatchID,
 	}, nil
--- a/internal/snell-rollcall/codec/payload_menu32.go
+++ b/internal/snell-rollcall/codec/payload_menu32.go
@@ -168,12 +168,12 @@ func DecodeMenuItem(b []byte) (MenuItem, error) {
 
 	rest := b[MenuItemSize:]
 	if len(rest) > 0 {
-		s, n := cString(rest)
+		s, n := CString(rest)
 		m.Text = s
 		rest = rest[n:]
 	}
 	if len(rest) > 0 {
-		m.Param, _ = cString(rest)
+		m.Param, _ = CString(rest)
 	}
 	return m, nil
 }
@@ -207,8 +207,8 @@ func (m MenuItem) ToFunc() (Func, error) {
 		MaxRange:  m.MaxRange,
 		Step:      uint16(m.Step),
 		DivScale:  m.DivScale,
-		Text:      truncateFixed(m.Text, MaxTextSize),
-		Param:     truncateFixed(m.Param, MaxTextSize),
+		Text:      TruncateFixed(m.Text, MaxTextSize),
+		Param:     TruncateFixed(m.Param, MaxTextSize),
 	}, nil
 }
 
--- a/internal/snell-rollcall/codec/payload_session.go
+++ b/internal/snell-rollcall/codec/payload_session.go
@@ -173,7 +173,7 @@ func DecodeWait(b []byte) (Wait, error) {
 		Mode:    Mode(binary.BigEndian.Uint16(b[2:4])),
 	}
 	if w.Mode.Has(ModeString) && len(b) > WaitSize {
-		w.Reason, _ = cString(b[WaitSize:])
+		w.Reason, _ = CString(b[WaitSize:])
 	}
 	return w, nil
 }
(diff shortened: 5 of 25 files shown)


by-rune:

PR #1021, opened 2026-09-07
feat(snell-rollcall): codec part 3 — complete the spec: file, time, log, stream, display, group

### LINKED ISSUE #1020: feat(snell-rollcall): codec part 3 — complete the spec: file, time, log, stream, display, group
Unit 4b of #1009, inserted before the consumer. Follows #1019.

## Why

A coverage check of the codec against the vendor headers found **15 wire
structures still unimplemented**. The consumer and provider cannot be
spec-complete on top of a codec that is not, so this closes the gap first.

Implemented today: `MESSAGE_STR`, `ROLLHEADER_STR`, `FULLADDRESS_STR`,
`VERSION_STR`, `ID_STR`, `STATUS_STR`, `DEVICEINFO_STR`, `CONNECT_STR`,
`TERMSESS_STR`, `CLEARSESS_STR`, `WAIT_STR`, `GETFSTAT_STR`,
`FUNCSTATUS_STR`, `SETMULTI_STR`, `DISP_STR`, `FUNC_STR`, `FUNCSTYLE_STR`,
`BLOCKHEADER_STR`, `GETNEXT_STR`, `MENUREQ_STR`, `MENUSIZE_STR`,
`MENUITEM_STR`, `GETVALUE_STR`, `VALUE_STR`.

Missing, and in this unit:

| Service | Structures | Size |
| --- | --- | --- |
| File | `FILE_STR`, `FILEINFOHDR_STR`, `MODULE_FILEINFO_STR` | 10, 10, 24 |
| Time | `TIME_STR`, `TIMECODE_STR` | 14, 6 |
| Logging | `LOGPACKET_STR` | 38 + text |
| Stream | `STREAMMODE_STR`, `STREAMHDR_STR` | 4, 8 + data |
| Display | `DISPLAYCAPS_STR`, `DRAWBITMAP_STR`, `DRAWTEXT_STR` | 12, 2 + data, 2 + text |
| Group | `SETGROUP_STR`, `GROUPPARAM_STR` | 2, 2 |
| Exec | `DOWNLOAD_STR` | 2 |

Plus the flag sets that go with them: `AttributeFlags`, `FileOpenFlags`,
`ErrorValues`, `TimeModeFlags`, `TimeCodeFlags`, `StreamModeFlags`, and the
display colour formats.

## Why the file service matters most

Two things in the audit depend on it and neither can be built without it.

The Control Panel will not render a producer without `TEMPLATE.ZIP` served
over `SV_FILE`. The menu service alone is not enough, which was established
against the live Control Panel 4.12.48.

And the router command set fetches every source and destination name from
files, referenced by `{filename, crc}` pairs. The layouts and the checksum
already landed in #1017; the transport for them is here.

## A note on one size

`MODULE_FILEINFO_STR` is a 10-byte header plus a 13-byte filename, which is
23 bytes of content. Under `#pragma pack(2)` the structure is padded to 24,
and the vendor writes `sizeof`, so 24 bytes go on the wire. This is the same
shape as the pad byte already documented on `GETNEXT_STR`. Encode 24, accept
either.

`ROUTEERROR_STR` is referenced by the packet-type table but never defined in
any shipped header. It is treated as an opaque payload and the gap is
recorded rather than guessed at.

## Acceptance

- `go test -race` green.
- 100% statement coverage, CI floor already in place for the package.
- A coverage matrix in `internal/snell-rollcall/docs/` listing every packet
  type and every structure against its implementation, so "is the spec
  covered" has an answer that can be checked rather than asserted.
- Expected bytes from the specification and the headers, never from this
  implementation.
- Still stdlib-only, no `dhs/*` import.

by-rune:
Closes #1020. Part of #1009. Inserted before the consumer, stacked on #1019.

## Why this exists

You asked to make sure the whole specification is supported. I checked the
codec against the vendor headers before starting the consumer, and it was not:
**fifteen wire structures were unimplemented**. Building consumer and producer
verbs on that would have produced a connector that looked complete and quietly
could not speak half the protocol.

## What was missing

| Service | Structures |
| --- | --- |
| File | `FILE_STR`, `FILEINFOHDR_STR`, `MODULE_FILEINFO_STR` |
| Time | `TIME_STR`, `TIMECODE_STR` |
| Logging | `LOGPACKET_STR` |
| Stream | `STREAMMODE_STR`, `STREAMHDR_STR` |
| Display | `DISPLAYCAPS_STR`, `DRAWBITMAP_STR`, `DRAWTEXT_STR` |
| Group | `SETGROUP_STR`, `GROUPPARAM_STR` |
| Exec | `DOWNLOAD_STR` |

All fifteen are now implemented, with the flag sets that go with them: file
attributes, open flags, errno values, seek origins, time and timecode modes,
stream modes, and the display colour formats.

## The file service is the one that unblocks real work

Two things from the audit depend on it and neither could be built without it.

The Control Panel **will not render a producer** without `TEMPLATE.ZIP` served
over `SV_FILE`. That was established against the live Control Panel 4.12.48;
the menu service alone is not enough.

And every router source and destination name is fetched from a file, named by
a `{filename, crc}` pair. The layouts and the checksum landed in #1017. This is
the transport for them.

## Three details worth calling out

**`MODULE_FILEINFO_STR` is 24 bytes, not 23.** Its content is a 10-byte header
plus a 13-byte name, but under `#pragma pack(2)` the structure is padded to an
even size and senders use `sizeof`. Same shape as the pad already documented on
`GETNEXT_STR`: encode 24, accept either.

**`TIME_STR` follows C.** Month is zero-based and the year counts from 1900. A
reader that misses this is a month and a century out, and the values look
plausible either way.

**Which parts of `TIME_STR` are meaningful is stated, not implied.** A unit
with no real-time clock sends uptime and leaves the date fields at zero.
Reading those as a date puts 1 January 1900 in an operator's log, so `Time()`
returns false unless the real-time flag is set.

## The coverage question now has a checkable answer

New: `internal/snell-rollcall/docs/spec-coverage.md`. Every packet type and
every structure against its implementation, the flag tables, the known gaps,
and what is deliberately out of scope with the reason.

It is **verified by tests rather than asserted**. `TestSpecCoverage` requires a
row for every one of the 72 packet types, a row naming the Go type for every
vendor structure, and a round trip through all 34 encoders and decoders. The
document cannot drift from the code without CI saying so.

## Two gaps, recorded rather than guessed

**`ROUTEERROR_STR` has no definition.** Packet type 38 refers to it in
`rc3comm.h` and no shipped header declares it. Type 38 is carried as an opaque
payload with the frame delivered intact, so a caller can inspect it. Inventing
a structure would be a guess presented as fact.

**The router sub-table offsets are inferred.** The Full Control document
numbers only its root table and describes the rest as "defined in a similar
fashion", so the offsets are read from that ordering. Unvalidated, because the
audit oracle answers NACK to command 100 and exposes no routing interface. A
test fails if the document stops saying so.

## Verification

```
ok  dhs/internal/snell-rollcall/codec          coverage: 100.0% of statements
ok  dhs/internal/snell-rollcall/codec/dtp      coverage: 100.0% of statements
ok  dhs/internal/snell-rollcall/codec/router   coverage: 100.0% of statements
ok  dhs/internal/snell-rollcall/session        coverage: 100.0% of statements
```

186 tests in the codec package. Every structure the vendor headers declare now
has a Go equivalent. Still stdlib-only with no `dhs/*` import, and building for
five OS/arch pairs with cgo disabled.

## Next

Unit 5, the consumer, on a codec that now covers the whole wire format.

### DIFF
--- a/internal/snell-rollcall/codec/payload_file.go
+++ b/internal/snell-rollcall/codec/payload_file.go
@@ -1,0 +1,325 @@
+package codec
+
+import (
+	"encoding/binary"
+	"fmt"
+)
+
+// Wire sizes of the file service structures (RC3FILE.H).
+const (
+	// FileSize is FILE_STR, the structure every file operation carries.
+	FileSize = 10
+
+	// FileInfoSize is FILEINFOHDR_STR, one directory entry's metadata.
+	FileInfoSize = 10
+
+	// FileNameSize is FILENAMESIZE: the fixed filename field in a directory
+	// entry. It is a DOS 8.3 name with its terminator.
+	FileNameSize = 13
+
+	// DirEntrySize is MODULE_FILEINFO_STR as it appears on the wire.
+	//
+	// The header plus the filename is 23 bytes of content, but the structure
+	// is compiled under #pragma pack(2) and so is padded to an even size, and
+	// senders use sizeof. The pad byte's content is undefined, exactly like
+	// the one in GetNext, and must never be reported as a deviation.
+	DirEntrySize = 24
+
+	// MaxFileName is the longest path the file service accepts, from
+	// RC3FILE.H MAX_FILENAME. Paths travel as NUL-terminated strings in the
+	// payload rather than in the fixed field, which is only used by directory
+	// listings.
+	MaxFileName = 131
+)
+
+// File open flags (RC3FILE.H FileOpenFlags). They mirror the C runtime's
+// open(2) flags, because that is what the vendor implementation passes them to.
+const (
+	OpenReadOnly  uint16 = 0x0000
+	OpenWriteOnly uint16 = 0x0001
+	OpenReadWrite uint16 = 0x0002
+	OpenAppend    uint16 = 0x0008
+	OpenCreate    uint16 = 0x0100
+	OpenTruncate  uint16 = 0x0200
+	OpenExclusive uint16 = 0x0400
+	OpenText      uint16 = 0x4000
+
+	// OpenBinary must be set for anything that is not line-oriented text.
+	// Without it the peer translates line endings, which corrupts a names
+	// file or a template archive silently.
+	OpenBinary uint16 = 0x8000
+)
+
+// File attribute flags (RC3FILE.H AttributeFlags).
+const (
+	AttrNormal   uint16 = 0x0000
+	AttrReadOnly uint16 = 0x0001
+	AttrHidden   uint16 = 0x0002
+	AttrSystem   uint16 = 0x0004
+	AttrVolumeID uint16 = 0x0008
+	AttrSubdir   uint16 = 0x0010
+	AttrArchive  uint16 = 0x0020
+
+	// AttrDOSFileTime says the time field is in MS-DOS date and time format
+	// rather than seconds. It is a flag about the encoding of another field,
+	// which is why it sits oddly high among the attribute bits.
+	AttrDOSFileTime uint16 = 0x1000
+	AttrRecord      uint16 = 0x2000
+	AttrWriteOnly   uint16 = 0x4000
+)
+
+// File service error values (RC3FILE.H ErrorValues). They are C errno numbers,
+// carried in the Extra field of a reply.
+const (
+	FileErrNoEntry     int16 = 2   // no such file or directory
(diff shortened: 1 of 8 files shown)


by-rune:

PR #1023, opened 2026-09-07
feat(snell-rollcall): consumer — the outbound connector, both generations

### LINKED ISSUE #1022: feat(snell-rollcall): consumer — the outbound connector, both generations
Unit 5 of #1009. Deliverable 1 of ADR-0025. Follows #1021.

## Scope

`internal/snell-rollcall/consumer/`, implementing `consumer.Protocol` and
registering as `rollcall`.

Built on the session layer from #1019 and the now spec-complete codec from
#1021, so every verb has a wire form to use and both generations are reachable.

| File | Contents |
| --- | --- |
| `plugin.go` | the factory, registration, and the plugin type |
| `connect.go` | dial, handshake, generation negotiation, sessions |
| `discover.go` | the device map and port list, into slots |
| `walk.go` | menu walk into objects, both generations, with the tree spans |
| `value.go` | get and set, both generations |
| `subscribe.go` | the back channel, into events |
| `file.go` | the file service: read a named file |
| `compliance_events.go` | the deviation catalogue |

## Mapping onto the neutral model (ADR-0022)

| RollCall | dhs |
| --- | --- |
| gateway unit, port 0 | Device / Frame |
| unit address | Device |
| port within a unit | Slot |
| `ID_STR` type id and version | Card identity |
| menu line | Object |
| command number | Object ID |
| menu style | ValueKind and access |

## Generation negotiation

The rule from the audit, and the one thing a client must not get wrong. The
peer advertises `SV_LONGSTR` in its service mask. A client that wants the
32-bit generation asks for it in the Call. Services are all-or-nothing, so a
peer that cannot supply the bit refuses the whole call and the client must
issue a second one without it rather than expect a partial grant.

The generation is a property of the session, not the device: the same unit
serves both, one session each.

## Menu model

A menu is a flat array with nested spans. On a container line the step field
is the span of the whole subtree, not the count of immediate children, so
rebuilding the tree is a pre-order walk that consumes that many entries.
Getting this wrong nests everything below the second level incorrectly.

Level gating is a substitution, not a removal: a line above the session user
level comes back as a hidden, disabled Data line named "Reserved". Line counts
are identical at every level, so the walker must compare flags.

## Acceptance

- `go test -race` green, 100% statement coverage, CI floor.
- Every verb in `docs/protocols/verb-tests.md` that the protocol supports.
- Both generations exercised, over a fake peer with an injected clock.
- `--label` and `--path` resolve cold, walking on a miss.
- Takes `plugin.Deps`; opens no socket except through the injected transport.
- Never imports the provider.

## Out of scope

The provider (unit 6), the dissector (unit 7), and the Ansible integration
(unit 8). Router verbs land with the matrix work once the command set can be
validated against a controller.

by-rune:
Closes #1022. Deliverable 1 of ADR-0025. Part of #1009 (unit 5 of 10).

Stacked on #1021.

## What it does

Implements `consumer.Protocol`, registers as `rollcall`, and speaks both wire
generations on the session layer from #1019 and the spec-complete codec from
#1021.

| RollCall | dhs |
| --- | --- |
| gateway unit | Device |
| port within a unit | Slot |
| a port's type id and version | Card identity |
| menu line | Object |
| command number | Object ID |
| menu style | ValueKind and access bits |

Verbs: connect, device info, slot info, walk, get, set, set-default,
subscribe, unsubscribe, display lines, device map, port list, read file,
stream file, list directory.

## Generation negotiation

The peer advertises long strings; this connector asks for them whenever it
sees the bit, because that generation is the only one that can express a
router's command space and its menu requests are stateless.

Services are all-or-nothing, so there is no partial grant to fall back to. A
peer that advertises the bit and then refuses a call asking for it gets a
second call without it, and **the fallback is recorded as a compliance
event**: the two statements cannot both be true, and which commands are
reachable depends on which one won.

## Two protocol behaviours pinned by tests

**A container's step is a span, not a child count.** Building the tree is a
pre-order walk that consumes that many following entries. Read as a child
count, everything below the second level nests in the wrong place and the
result still looks like a tree. The test uses a menu where the two readings
put one line in different places.

**Level gating substitutes rather than removes.** A line above the session's
user level comes back as a hidden, disabled Data line named "Reserved". Line
counts are identical at every level, so a client comparing counts sees no
difference. Detected and reported.

## The file service earns its place

Two things depend on it and neither works without it. The Control Panel will
not render a device without reading `TEMPLATE.ZIP` from it, and a router
publishes every source and destination name in a file.

It opens its own session, because services are negotiated together and
all-or-nothing: asking for the file service on the control session would make
a unit that lacks one uncontrollable.

## Two defects found by the tests

**Decoded payloads aliased the reader's buffer.** Pushes wait in a queue until
the application takes them, so by the time a listener ran, every queued push
carried the bytes of whichever frame arrived last. Three different values
arrived as three copies of the third, and nothing about that fails loudly. The
link now copies each frame before dispatch; regression tests cover both the
push queue and the announcement queue.

**The file service's fields change meaning between request and reply.** The
vendor server says so outright:

```c
/* NOTE that meaning of rOffset and rExtra differs between SP_FILEREAD and SP_RETFILEREAD. */
/* - in SP_FILEREAD:    rOffset = offset into file, rExtra = number of bytes to read. */
/* - in SP_RETFILEREAD: rOffset = number of bytes actually read, rExtra = error code   */
```

And for open: `OpenModeFlags = FileSpecIn->rExtra`.

The codec had both wrong. Open flags were being sent in the offset, and the
read count was a "seek origin" that does not exist in this protocol. A read
would have asked for zero bytes every time and reported an empty file rather
than an error. Corrected, with the vendor source cited in the code, and the
open reply's block size is now honoured.

## Verification

```
ok  dhs/internal/snell-rollcall/consumer   coverage: 100.0% of statements
```

91 tests against a **fake RollCall device that speaks the wire format**, not a
mock of the session layer. A failure therefore means the bytes are wrong, not
that an expectation was written differently from the code. It runs over a pipe
with an injected clock: no socket, no port, no sleeping.

Every verb is exercised three ways: against a device that refuses it, against
one that answers with an undecodable payload, and in both generations. Five
consecutive runs, no flake.

## Coverage

**100% of statements, locked as a CI floor**, in line with every other package
in this connector.

Getting there closed twenty reachable branches and removed two duplicated
guards. The removals are the interesting half: `locate` asked for the session
after the menu walk that needs one anyway, leaving a path reachable only by a
disconnection landing between two calls; and `set16` checked whether its text
encoded after truncating that text to fit, which cannot fail and obscured the
fact that truncation is what the older generation requires.

One behaviour a new test pins: when a read reply understates how many bytes it
carries, the count is what the client believes and advances by, so the
declined byte is asked for again and the file still arrives whole. Taking the
payload while advancing by the count would duplicate a byte per block and
corrupt the middle of an archive.

## Also here

Registered in `cmd/dhs/main.go` and the frozen CLI-surface golden file updated,
which is the intended step when a protocol joins the registry:

```
+rollcall port=2050  Snell RollCall over IPShare, 16-bit and 32-bit generations
```

## Next

Unit 6, the provider, which the Control Panel already rendered once from a
throwaway spike during the audit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,12 +172,7 @@ jobs:
           check internal/snell-rollcall/codec/dtp 100
           check internal/snell-rollcall/codec/router 100
           check internal/snell-rollcall/session 100
-          # The consumer is the one package below the 100% standard. What
-          # remains is error propagation whose reachability depends on a
-          # disconnection landing inside a specific window; it rises as those
-          # branches are either exercised or shown to be duplicated guards and
-          # removed. The floor stops it slipping meanwhile.
-          check internal/snell-rollcall/consumer 96
+          check internal/snell-rollcall/consumer 100
           check internal/transport/ws 100
           check internal/transport/http 100
           check internal/auth 100
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,6 +172,12 @@ jobs:
           check internal/snell-rollcall/codec/dtp 100
           check internal/snell-rollcall/codec/router 100
           check internal/snell-rollcall/session 100
+          # The consumer is the one package below the 100% standard. What
+          # remains is error propagation whose reachability depends on a
+          # disconnection landing inside a specific window; it rises as those
+          # branches are either exercised or shown to be duplicated guards and
+          # removed. The floor stops it slipping meanwhile.
+          check internal/snell-rollcall/consumer 96
           check internal/transport/ws 100
           check internal/transport/http 100
           check internal/auth 100
--- a/cmd/dhs/main.go
+++ b/cmd/dhs/main.go
@@ -46,6 +46,7 @@ import (
 	_ "dhs/internal/osc/consumer"
 	_ "dhs/internal/probel-sw02p/consumer"
 	_ "dhs/internal/probel-sw08p/consumer"
+	_ "dhs/internal/snell-rollcall/consumer"
 	_ "dhs/internal/tsl/consumer"
 
 	// Provider plugins — blank imports register with internal/provider.
--- a/cmd/dhs/testdata/golden/cli-surface.txt
+++ b/cmd/dhs/testdata/golden/cli-surface.txt
@@ -76,6 +76,7 @@ osc-v10  port=8000  Open Sound Control 1.0 — UDP + TCP/length-prefix; core typ
 osc-v11  port=8000  Open Sound Control 1.1 — UDP + TCP/SLIP; adds T/F/N/I + arrays
 probel-sw02p port=2002  Probel SW-P-02 matrix controller (TCP)
 probel-sw08p port=2008  Probel SW-P-08 / SW-P-88 matrix controller (TCP)
+rollcall port=2050  Snell RollCall over IPShare, 16-bit and 32-bit generations
 tsl-v31  port=4000  TSL UMD v3.1 — 18-byte UDP frame (binary tallies, no colour)
 tsl-v40  port=4000  TSL UMD v4.0 — v3.1 extended with XDATA + CHKSUM (2-bit colour per LH/Text/RH per display)
 tsl-v50  port=8901  TSL UMD v5.0 — UDP (≤2048 B) or TCP with DLE/STX wrapper; multi-DMSG; UTF-16LE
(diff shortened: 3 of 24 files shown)


by-rune:

PR #1025, opened 2026-09-07
feat(snell-rollcall): provider — serve a tree as a gateway, both generations

### LINKED ISSUE #1024: feat(snell-rollcall): provider — serve a tree as a RollCall device, both generations
Unit 6 of #1009. Deliverable 2 of ADR-0025. Follows #1023.

## Scope

`internal/snell-rollcall/provider/`, implementing `provider.Provider` and
registering as `rollcall`.

It serves a canonical tree as a RollCall gateway: a frame with slots, each
slot a card with its own menu and values, discoverable and controllable by the
vendor's own Control Panel.

## The session layer needs a server side

`session` today is the calling side: it opens sessions and drives requests.
A provider is the same machine with the roles flipped, which is what the audit
said and what the vendor's own `Core/` is organised around, so this unit adds
the answering side to that package rather than duplicating it:

- an optional handler on a link, so inbound frames that match no client
  session reach the provider instead of the lossy announcement channel
- accepting an inbound Call: allocate our index, register the session, take
  the caller's index as sticky, answer with the acknowledgement
- routing inbound requests on an established session to the handler

## What the provider serves

| Service | Messages |
| --- | --- |
| identity | GetDevInfo, GetID, GetStat, Iam announcements |
| map and ports | GetLocDevMap, GetDevList |
| menus, 16-bit | GetFunc, GetNextPkt |
| menus, 32-bit | GetMenuCount, GetMenuItem |
| control, 16-bit | GetFStat, SetParam, RepFChg |
| control, 32-bit | GetValue, SetValue |
| display | GetDispData, DispData |
| file | FileDir, FileOpen, FileRead, FileClose |
| link | Call, Term, KeepAlive, BkChnReady |

Both generations from one tree, chosen per session from the caller's own
service mask, because the same unit serves both.

## TEMPLATE.ZIP is not optional

The Control Panel will not render a device without reading `TEMPLATE.ZIP` from
it over the file service. That was established live against Control Panel
4.12.48 during the audit: the menu service alone is not enough. The provider
generates and serves one.

## Where the router document changes things

"Full RollCall Control of Routers" (V1, September 2013) settles how a router
is exposed, and it is not how a card is:

- routing and names are served as **control variables, not menu lines**, so a
  router node has no menu to walk
- the template file defines the well-known command numbers a client uses
- names are bulk-loaded from files, because a client rendering an 18k square
  level from menu lines "sits there attempting to render the template for
  longer than I am prepared to wait"
- a refused route returns NACK **with a text string** the client displays
  before reverting its own value
- protects need a unique panel id and a master flag, both of which the
  command set already carries

That shapes the provider's value path even for cards, and it is what the
router unit will build on.

## Acceptance

- `go test -race` green, **100% statement coverage**, CI floor.
- Both generations served from one tree, exercised by the consumer from #1023
  driving it over a pipe.
- Every message above answered, with the right refusal for the ones a given
  tree cannot serve.
- `provider.Provider` satisfied; takes `plugin.Deps`, binds only through the
  injected transport.
- Never imports the consumer.

## Out of scope

The router command set as a served interface (its own unit, with the matrix
model), the dissector, and the Ansible integration.

by-rune:
Closes #1024. Sixth unit of the RollCall connector, stacked on #1023.

## What it does

`dhs producer rollcall serve --tree tree.json` serves a canonical tree as a
Snell RollCall gateway. The root is the frame, each child of the root is a card
slot, and everything below a slot is that card's menu. Port zero is the gateway
itself — cards start at port one, because a client's first message asks port
zero "what are you?" and a card there would answer that with the wrong identity.

Which wire generation a client gets is decided by the service mask in its
`SP_CALL`, not by what we are. One model answers both, with the same command
numbers and the same labels; a number that will not fit the older generation is
served as a disabled placeholder rather than truncated, because a truncated
command number addresses a different command.

Served: menus in both forms (spans nested by subtree, not child count), values
with a write clamped and the stored value in the reply, back-channel pushes
queued per subscriber and encoded per generation, the file service with a
generated `TEMPLATE.ZIP` (the vendor Control Panel draws nothing without it),
display lines including the two negative priorities, and the device map.
Anything a client gets wrong is absorbed and counted as a compliance event.

## Two real defects fixed on the way

- **`Session.Close` deadlocked on a dead socket.** `Term` held a `sync.Once`
  while sending; a failed send closes the link, closing a link closes its
  sessions, and that re-entered the same `Once` on the same goroutine. Now a
  flag, with a regression test that fails by timeout without the fix.
- **`Serve` deadlocked when its listener failed on its own** — the accept loop
  waited for the goroutine waiting for the listener to close.

Two smaller ones came out of driving the CLI against it end to end: the consumer
left the wire-level `SlotStatus` unset, so `dhs consumer rollcall info` reported
every present card as `no_card`; and a double-close guard in `Disconnect` was
unreachable by construction.

## Coverage

| package | before | after |
|---|---|---|
| `snell-rollcall/provider` | — | **100.0%** |
| `snell-rollcall/session` | 100% (server side untested) | **100.0%** |
| `snell-rollcall/consumer` | 100% | **100.0%** |

`internal/snell-rollcall/provider 100` added to the CI floors. The provider is
driven by a real client over a pipe — a `session.Link` speaking the wire format,
deliberately **not** the consumer plugin, because a provider and a consumer that
share an assumption agree with each other and with nothing else. Nothing sleeps.

## Loopback check (Tier 4, after the emulator and the device per ADR-0025)

```
dhs producer rollcall serve --tree bin/trees/our-dm/tree.json --port 25051 &
dhs consumer rollcall info 127.0.0.1:25051     # 7 slots, all present
dhs consumer rollcall walk 127.0.0.1:25051 --slot 1
dhs consumer rollcall set  127.0.0.1:25051 --slot 2 --id 1 --value "CAM 4"
dhs consumer rollcall get  127.0.0.1:25051 --slot 2 --id 1   # "CAM 4"
```

A write to a read-only line comes back `NACK (command 1 is read-only)`, which is
the access gating working rather than a failure.

## Not in this unit

No router node: a RollCall router serves routing and names as control variables
rather than menu lines, which is a different model and the next unit. No file
writes, and no logging/streaming/drawing services — they are advertised by
neither the gateway nor its cards, so a call for one is refused outright.

Docs: `internal/snell-rollcall/docs/provider.md`.


by-rune:

PR #1027, opened 2026-09-07
fix(snell-rollcall): reach every node of a real controller, ask for what it has

### LINKED ISSUE #1026: fix(snell-rollcall): consumer cannot reach a real controller — services and node addressing
Found by running the vendor Centra controller (`assets/Emulator/RouterSimulator/Router2.zip`,
`CentraController.exe` as a Sirius 800) and driving it with our own consumer.
Full measurements in `internal/snell-rollcall/docs/oracle-centra.md`.

## Two defects, both fatal against this controller

**1. We ask for a service the peer does not have.**
`sessionServices()` always includes `SV_DISPLAY`. This controller advertises
`Menus|Control|File|Map|LongStr` and refuses any call including Display —
bracketed by the identical call without it succeeding immediately before and
after. So `dhs consumer rollcall info 127.0.0.1:2057` cannot open a session at
all:

```
error: rollcall: call 0000-08-00:0FF: rollcall: call: NACK
```

The requested mask must be intersected with what the peer advertised.

**2. We address ports on one unit; a real gateway's nodes are units.**
The Centra's device list has fifteen entries and they differ by *unit*, not by
port: Matrix 1 at `0000-11-00`, Matrix 2 at `0000-12-00`, TIELINES at
`0000-80-00`, XY Panel at `0000-81-00`, cards at `0000-41-00`…`0000-66-00`.
`p.session(ctx, port)` only ever varies the port on the gateway's own unit, so
fourteen of the fifteen nodes are unreachable and a slot read times out.

Sessions have to be keyed and addressed by the full device address taken from
the enumeration, not by a port number we assumed.

## Acceptance

- Session masks intersect with the peer's advertised services; a peer without
  Display is usable.
- Nodes are addressed by the address the device list gave them. Where the list
  is one unit's ports (a frame) the slot number is the port; where it is a set
  of units (a controller) the slot number is the unit. Which rule is in force
  is logged once, not guessed silently per call.
- `dhs consumer rollcall info` and `walk` work against the vendor Centra.
- 100% statement coverage kept, CI floor unchanged.

Prerequisite for router control (the routing interface lives on the XY Panel
node, which is unreachable today).

by-rune:
Closes #1026. Stacked on #1025.

Found by running the vendor Centra controller — `assets/Emulator/RouterSimulator/Router2.zip`,
`CentraController.exe` as a **Sirius 800** — and pointing our own consumer at
it. Measurements: `internal/snell-rollcall/docs/oracle-centra.md`.

## Before

```
$ dhs consumer rollcall info 127.0.0.1:2057
error: rollcall: call 0000-08-00:0FF: rollcall: call: NACK
```

## Two defects, both ours

**We asked for a service the peer does not have.** The mask always included
`SV_DISPLAY`; this controller advertises `Menus|Control|File|Map|LongStr` and
refuses the whole call, because services are all-or-nothing. Bracketed: the
identical call without Display succeeds immediately before and after. The mask
is now intersected with what the peer advertised, **per node** — each device-list
entry carries its own services and they differ between the matrices, the panel
node and the gateway.

**We addressed a slot as a port on the gateway's unit.** This controller's
nodes differ by *unit*: Matrix 1 at `0000-11-00`, TIELINES at `0000-80-00`, and
the XY Panel that serves the routing interface at `0000-81-00`. Fourteen of
fifteen were unreachable, and a slot read timed out rather than failing. A slot
is now a position in the device's own enumeration with the address the device
gave it, and that address appears in the slot's identity.

Also: the device map gets **its own session**. A request outside a session's
negotiated services is answered `SP_INVSESS`, so enumerating on the control
session failed. Asked for alone, without the long-string bit — which this
controller refuses in combination and which buys nothing on a fixed-width
structure — with the control session as the fallback.

## After

```
$ dhs consumer rollcall info 127.0.0.1:2057
slots        15
  slot  0   status=present    online=true
  …
  slot 14   status=present    online=true

$ dhs consumer rollcall walk 127.0.0.1:2057 --slot 14
slot 14 — 3 objects
```

Three objects because slot 14 is the XY Panel: a router serves routing and
names as control variables, not menu lines. That is the design document's whole
point and why a menu walk finds nothing on a router.

## The same run validated `codec/router`

Every offset in every sub-table was read individually off the controller, and
its content identifies the field. All four table sizes confirmed (matrix 15,
level 12, source 3, destination 8); `CategoryBase = MatrixBase + N × MatrixStep`
checks out at 151 = 121 + 2×15; source-pin packing confirmed; and the
names-file checksum reproduces the controller's published values **exactly**
for three different files. The package comment no longer calls those offsets
unvalidated.

A live route was set and read back, which established two things a client must
get right: the reply to a set carries the **previous** pin (the new one arrives
as a back-channel push), and the tally **follows tielines** to the upstream
source, so reading back what you wrote is not a valid expectation.

The controller reports interface version **13**; our specification copy
documents 12.

## Coverage

100% of statements in all six packages, unchanged floors. Two test harness
changes were needed and both make the tests more honest: the fake device now
advertises per-node services in its device list (as the vendor does), and the
tests that force a session race now enumerate first, so the gate holds the
racing calls rather than the enumeration that precedes them.

Verified against both the vendor Centra and our own provider from #1025.

### DIFF
--- a/internal/snell-rollcall/codec/router/commands.go
+++ b/internal/snell-rollcall/codec/router/commands.go
@@ -15,12 +15,21 @@
 //
 // The root table is pinned by the specification, which states outright that
 // CMD_INTERFACE_VERSION is command 100 and the rest follow in order. The
-// per-matrix, per-level and per-destination tables are documented as "defined
-// in a similar fashion" with their commands listed in order but no numbers
-// given, so their offsets are read from that ordering. That reading has not yet
-// been checked against a router controller: the audit oracle answered NACK to
-// command 100, so it exposes no routing interface at all. Treat the sub-table
-// offsets as unvalidated until a real controller confirms them.
+// per-matrix, per-level and per-entity tables are documented as "defined in a
+// similar fashion", with their commands listed in order but no numbers given,
+// so their offsets were read from that ordering.
+//
+// Those offsets have since been measured against the vendor Centra controller
+// running as a Sirius 800, which serves the routing interface on its XY Panel
+// node: every command in every sub-table was read individually and its content
+// identifies which field it is, and the four table sizes here — matrix 15,
+// level 12, source 3, destination 8 — are what that controller publishes. The
+// names-file checksum below reproduces the controller's own values exactly.
+// See internal/snell-rollcall/docs/oracle-centra.md for the measurements.
+//
+// The controller advertises interface version 13; the specification copy we
+// hold documents 12. Commands are only ever added, so everything named here
+// exists, but whatever 13 introduced is not described anywhere we can read.
 //
 // This package is stdlib-only and imports nothing from dhs (ADR-0006).
 package router
--- a/internal/snell-rollcall/consumer/compliance_events.go
+++ b/internal/snell-rollcall/consumer/compliance_events.go
@@ -50,6 +50,12 @@ const (
 	// cached walk does.
 	EventUnsolicitedCommand = "rollcall_unsolicited_command"
 
+	// EventUnlistedNode means a slot was addressed that the device's own
+	// enumeration does not mention. It is addressed anyway: a gateway ages a
+	// map entry out after sixty seconds of silence, so a node missing from the
+	// list is one that has gone quiet rather than one that never existed.
+	EventUnlistedNode = "rollcall_unlisted_node"
+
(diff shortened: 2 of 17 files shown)


by-rune:

PR #1029, opened 2026-09-07
feat(snell-rollcall): router control — the Full Control command set client

### LINKED ISSUE #1028: feat(snell-rollcall): router control — the Full Control command set client
A RollCall router has no menu. Routing, names, protects and salvos are served
as **control variables** in a flat 32-bit command space whose numbers a client
works out by arithmetic from a root table at command 100. A menu walk against
one finds three lines and nothing useful, which is what our consumer does today.

Measured behaviour, from the vendor Centra as a Sirius 800:
`internal/snell-rollcall/docs/oracle-centra.md`. The arithmetic and the file
formats are already implemented and now validated in `codec/router`.

## Scope

Consumer-side router client:

- **Find it.** Probe command 100 on each enumerated node. It answers on the
  panel node (XY Panel on the Centra) and is refused everywhere else, so
  probing is safe.
- **Read the model.** Root table → matrices → levels → sources and
  destinations, all by arithmetic from the bases and steps the controller
  publishes.
- **Names in bulk.** Every name set is a file plus a checksum; fetch over the
  file service, verify with `NamesFile.CRC()`, cache by checksum. Sixty-five
  thousand names as individual commands is not an option.
- **Read and set crosspoints.** `CMD_DEST_ROUTED_SRC` carries Data Transfer
  Params: a packed source pin on read, a uint array on write.
- **Protects.** `CMD_DEST_PROTECT_STATE`, including the protect id and the
  master-panel bit.
- **Tally.** The new crosspoint arrives as a back-channel push, not in the
  reply to the set.

## Two behaviours a client must get right

- **The reply to a set carries the previous pin.** The result code is appended
  and says whether it worked; the new value comes on the back channel. Trusting
  the reply shows a stale crosspoint.
- **The tally follows tielines.** Setting source 7 can read back as matrix 2
  source 1, because the pin reports the final upstream source. Reading back
  what you wrote is not a valid expectation.

## Acceptance

- Discovery, model read, names bulk-load, read/set crosspoint, protect, tally.
- Unit tests against a fake router device at 100% statement coverage.
- Verified against the vendor Centra simulator, both matrices and every level.
- CLI verbs follow in a separate unit.

Needs #1026 (the routing interface is on a node our consumer could not reach).

by-rune:
Closes #1028. Stacked on #1027.

A RollCall router has no menu. Routing, names, protects and salvos are **control
variables** in a flat 32-bit command space whose numbers a client works out by
arithmetic from a root table at command 100. Walking a router's menu finds three
lines and nothing useful, which is what our consumer did until now.

Built and checked against the vendor Centra controller running as a **Sirius
800**. Measurements: `internal/snell-rollcall/docs/oracle-centra.md`.

## What it does

```go
r, _ := p.FindRouter(ctx)                       // probes command 100 on each node
n, _ := p.LevelNames(ctx, r, 1, 1, 32)          // bulk file, checksum-verified
p.WatchRoutes(ctx, r, func(x Crosspoint) {…})   // tally
p.SetRoute(ctx, r, 1, 1, 1, src, 0)             // make a route
p.Protect(ctx, r, 1, 1, 1)                      // who holds it
p.FireSalvo(ctx, r, 2)
```

Live, against the vendor controller:

```
router on slot 14 v13: 2 matrices, 4 salvos
m2/l3 names: mtx02src000001 … mtx02src000010 | mtx02dst000001 … mtx02dst000010
tally m2/l3/d2 <- m2/l3/s2
tally m2/l3/d2 <- m2/l3/s5
level 2/3 crosspoints: d1<-s1 d2<-s5 d3<-s1 d4<-s1 …
```

## Three behaviours a client gets wrong without being told

- **The reply to a set carries the previous crosspoint.** The result code says
  whether it worked; the new pin arrives afterwards as a back-channel push. A
  client that reads the reply's pin shows a stale crosspoint.
- **The tally follows tielines.** Setting source 7 can read back as matrix 2
  source 1: the pin is the *final upstream* source. Reading back what you wrote
  is not a valid expectation.
- **Enabling the back channel does not replay current state.** The
  specification says it asks for what has already changed; this controller sends
  nothing until something moves. A panel reads once with `Routes` and keeps it
  current from `WatchRoutes`.

## Names in bulk, and when they cannot be

Every name set is a file plus a checksum computed from the names rather than the
bytes, so a cached copy is trusted without fetching. But this controller names
files in **its own** filesystem, and each node's file service is rooted at that
node's directory, so `RC_Files\…` cannot be opened from anywhere. The client
falls back to one command per name — which works, and is slow — recording
`rollcall_names_file_unreadable` so the slowness is not a mystery. On a
sixty-five-thousand-source level that fallback is very slow, which is precisely
why the file mechanism exists.

## One codec fix

**Directory entries are variable-length.** `MODULE_FILEINFO_STR` declares a
thirteen-byte name field and no server sends it: the header, the name, its
terminator, nothing more — twelve bytes for `.`, twenty-three for
`TEMPLATE.ZIP`. Our decoder demanded the declared width and could read *none* of
a real listing. The captured bytes are now the test.

## Coverage

100% of statements in all six packages. The fake controller in the tests is laid
out by the same arithmetic as the real one — base 121, step 15, everything below
found from what the level above published — so a reader that gets an offset
wrong fails against it for the same reason it would fail against a Sirius.

CLI verbs follow in a separate unit.

### DIFF
--- a/internal/snell-rollcall/codec/payload_file.go
+++ b/internal/snell-rollcall/codec/payload_file.go
@@ -364,15 +364,23 @@ func (d DirEntry) AppendTo(dst []byte) ([]byte, error) {
 
 // DecodeDirEntry reads a directory entry.
 //
-// Twenty-three bytes are accepted as well as twenty-four: the content is 23
-// and a peer that trims the structure's padding is not wrong.
+// The name is read as whatever follows the header rather than as a fixed
+// thirteen-byte field, because that is what servers send. The structure
+// declares the field, but the vendor's own controller sends the header
+// followed by the name and its terminator and nothing more: an entry for "."
+// arrives in twelve bytes, one for "TEMPLATE.ZIP" in twenty-three. Demanding
+// the declared width makes every short entry unreadable, which is most of them.
+//
+// A name that fills its field without a terminator is accepted too, which is
+// what a server that does send the full width produces for a thirteen-character
+// name.
 func DecodeDirEntry(b []byte) (DirEntry, error) {
-	const content = FileInfoSize + FileNameSize
-	if err := need(b, content, "DirEntry", ""); err != nil {
+	if err := need(b, FileInfoSize+1, "DirEntry", ""); err != nil {
 		return DirEntry{}, err
 	}
+	name, _ := CString(b[FileInfoSize:])
 	return DirEntry{
 		Info: fileInfoAt(b),
-		Name: fixedString(b[FileInfoSize:content]),
+		Name: name,
 	}, nil
 }
--- a/internal/snell-rollcall/codec/payload_file_test.go
+++ b/internal/snell-rollcall/codec/payload_file_test.go
@@ -296,25 +296,61 @@ func TestDirEntry_PaddedToTwentyFour(t *testing.T) {
 	}
 }
 
-// TestDirEntry_AcceptsTheTrimmedForm covers a peer that sends the 23 bytes of
-// content without the structure's padding. It is not wrong, and the entry must
-// still decode.
-func TestDirEntry_AcceptsTheTrimmedForm(t *testing.T) {
+// TestDirEntry_AcceptsWhatServersActuallySend covers the lengths a directory
+// listing arrives in.
+//
+// The structure declares a thirteen-byte name field, and servers do not send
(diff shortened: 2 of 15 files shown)


by-rune:

PR #1030, opened 2026-09-07
feat(snell-rollcall): router verbs — router, route, tally

by-rune:
Stacked on #1029, which built the router client with no way to reach it from a
command line.

```
dhs consumer rollcall router <host>                       what it publishes
dhs consumer rollcall route  <host> --matrix 1 --level 1 --dest 4 [--source 9]
dhs consumer rollcall tally  <host> --matrix 1 --level 1 [--names]
```

Live, against the vendor Centra as a Sirius 800:

```
$ dhs consumer rollcall router 127.0.0.1:2057
router       <not set>
slot         14 (0000-81-00)
interface    version 13
salvos       4
devices      100

MATRIX         LEVEL  NAME          SOURCES  DESTINATIONS
1 R1 Matrix 1  1      R1M1 Level 1  10       10
1 R1 Matrix 1  2      R1M1 Level 2  10       10
2 R1 Matrix 2  1      R1M2 Level 1  10       10
2 R1 Matrix 2  2      R1M2 Level 2  10       10
2 R1 Matrix 2  3      R1M2 Level 3  10       10

$ dhs consumer rollcall tally 127.0.0.1:2057 --matrix 2 --level 1 --names
m2/l1/d1    <- m2/l1/s1   mtx02dst000001 <- mtx02src000001
…
--- following changes; Ctrl+C to stop
```

## Each verb is shaped by something measured, not by the API

- **`route` waits for the tally.** Neither the reply nor an immediate read says
  what happened: the reply carries the crosspoint from *before* the set, and the
  controller applies asynchronously. Using the verb against the Centra is how
  that surfaced — it printed the previous source until it subscribed first.
- **`tally` reads the level before following it**, because enabling the back
  channel does not replay current state on a real controller. Subscribe-only
  shows nothing until something moves.
- **`router` probes when no `--slot` is given**, because nothing in a device
  list says which node serves the interface.

A route made from one process shows up in another's tally, which is the check
that matters for a panel.

The CLI surface golden gains the three verbs; `cmd/dhs` has no coverage floor,
and the new pure helpers plus argument validation are tested.

### DIFF
--- a/cmd/dhs/cli_surface_golden_test.go
+++ b/cmd/dhs/cli_surface_golden_test.go
@@ -53,7 +53,7 @@ var genericVerbs = []string{
 	"info", "walk", "tree", "get", "set", "inc", "dec", "reset", "ensure",
 	"watch", "export", "import", "extract", "diff", "convert", "discover",
 	"matrix", "usage", "replace", "invoke", "stream", "profile", "diag",
-	"validate", "health", "status", "bench",
+	"validate", "health", "status", "bench", "router", "route", "tally",
 }
 
 // emberplusOnlyVerbs / acp2OnlyVerbs are captured under their owning protocol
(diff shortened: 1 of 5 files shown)


by-rune:

PR #1031, opened 2026-09-07
feat(snell-rollcall): Wireshark dissector, validated against a live capture

by-rune:
Stacked on #1030. Every plugin ships a dissector (root `CLAUDE.md`), and this
protocol needs one more than most: two wire generations on one connection, a
router command space whose numbers mean nothing without arithmetic, and a reply
that carries the value from *before* the write. Reading a capture by eye is not
practical.

## What it decodes

- The transmission header with TCP reassembly — several messages in one
  segment, one message across segments.
- Addressing: net / unit / port / session index, with the unconnected index
  named rather than printed as 255.
- **All 72 packet types.**
- Payloads for everything a live device sends: sessions, identity, menus in
  both generations, values in both generations, displays, files, device maps,
  block transfers.
- The routing interface: root commands by name, Data Transfer Params decoded,
  crosspoints as `m1/l1/s6` rather than five bytes of varint.

Info column example, straight from the fixture:

```
RETVALUE       0000-81-00:48 → 0000-00-00:16  cmd=316 16842758 0 pin=m1/l1/s6
RETVALUE [back] 0000-81-00:48 → 0000-00-00:16  cmd=316 16842756 pin=m1/l1/s4
```

That is the whole "the reply carries the previous value, the new one arrives on
the back channel" behaviour, visible in two lines.

## The capture paid for itself immediately

Written against a live capture of our client driving the vendor Centra as a
Sirius 800. Four defects surfaced from garbled Info columns:

- the version field read as two words instead of four bytes;
- the block header's spare byte, which made a fifteen-node map read as zero;
- the menu item's command offset, off by two, which turned `"RETURN"` into
  `"RN"`;
- **and one in the consumer**: the Centra's input cards answer command 100 with
  their model number *as a string* rather than refusing it, so router discovery
  accepted a card as a router and only survived because the next command
  happened to be refused. The interface version is now required to be a number
  in a plausible range, with a test for each way a node can answer wrongly.

## Regression test

`tests/fixtures/snell-rollcall/centra-sirius800.pcapng` — 90 KB, 23 message
types — plus an integration-tagged replay test asserting the dissector reads
all of it with **no expert warnings**: the map of fifteen nodes, the routing
commands by name, crosspoints as source pins, and the back-channel tally told
apart from the reply. Skips where tshark is not installed.

```
go test -tags integration ./internal/snell-rollcall/integration/... -run DissectorReplay
```

`docs/wireshark.md` gains the dissector row and its display filters.


by-rune:

PR #1032, opened 2026-09-07
test(snell-rollcall): Ansible integration play, three tiers

by-rune:
Stacked on #1031. ADR-0025 deliverable 3, Ansible-only as the standing rule.

```
# loopback only
ansible-playbook -i inventory/hosts.ini playbooks/snell-rollcall-integration.yml

# plus the vendor Centra emulator
ROLLCALL_SIM_HOST=127.0.0.1 ROLLCALL_SIM_PORT=2057 ansible-playbook ...

# plus a live frame / Centra / Sirius, read-only
ROLLCALL_TEST_HOST=10.6.250.105 ansible-playbook ...
```

## The tier order is the argument

1. **Vendor emulator** — the only oracle that can disagree with us. It is the
   vendor's own code, speaks both generations, serves a routing interface.
   Read-only unless the operator nominates a crosspoint via `ROLLCALL_SIM_ROUTE`.
2. **Live device** — read-only, always. Nothing in a test writes to a real router.
3. **Loopback** — last, because two halves of one codebase agreeing proves the
   least. A regression net, not an oracle. Ephemeral producer torn down in
   `always`, so run-twice leaves nothing behind.

## Assertions are behaviours, not exit codes

That the controller enumerated its nodes; that the routing interface published
matrices and levels; that a crosspoint names a matrix, a level and a source;
that a value carries the unit its tree declared; that a write out of range
comes back **clamped**; that a read-only object **refuses** rather than
pretending. Every `fail_msg` says what the protocol does, so a red run is a
diagnosis rather than a puzzle.

## Fixture

`internal/snell-rollcall/testdata/exports/frame_tree.json` — two cards, with a
scaled real (`gain`, factor 10, dB), a checkbox, a string and a read-only
integer. Enough to exercise every mapping that differs.

## One defect found by building it

A card port advertised `Menus|Control|Display` and not the long-string bit, so
a client negotiated the **16-bit** generation against a card that serves both —
visible as `SETPARAM` where `SETVALUE` belonged. The provider now advertises
what it actually serves, which is what the vendor's own cards do.

Every assertion in the play was run by hand against the built binary first;
`ansible-playbook` is not installed on this machine, so the play itself is
unrun here and the YAML is validated only by parse.

### DIFF
--- a/ansible/playbooks/snell-rollcall-integration.yml
+++ b/ansible/playbooks/snell-rollcall-integration.yml
@@ -1,0 +1,355 @@
+---
+# Snell RollCall integration CONTRACT TEST, per ADR-0025 deliverable 3:
+# Ansible is the exclusive integration / verify / deploy driver — no shell
+# scripts in another language. This play drives the REAL dhs CLI verbs against
+# real servers, not `go test`.
+#
+# Three tiers, in the order ADR-0025 puts them:
+#
+#   1. VENDOR EMULATOR (gated on ROLLCALL_SIM_HOST): the Snell Centra
+#      controller from internal/snell-rollcall/assets/Emulator/RouterSimulator.
+#      It is the oracle that matters — it is the vendor's own code, it speaks
+#      both generations, and it serves a routing interface. Read-only verbs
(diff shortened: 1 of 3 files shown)


by-rune:

PR #1033, opened 2026-09-07
docs(snell-rollcall): the operator's set — README, runbook, consumer, oracle

by-rune:
Four documents an operator reads in order: what the connector is, what to run and what the answers mean, how the outbound side behaves, and what the vendor controller actually did when we measured it.

## The runbook is the load-bearing one

Bring-up in the order where each step tells you something the next one needs; the three usual causes of a refused session; Wireshark filters that answer a question rather than showing everything; the compliance-event table.

## The oracle document gains the knob that decides what the emulator is

Our first attempt to characterise `CENTRA_SIMULATED_ROUTER` got it wrong twice over. It is written down that way, because both mistakes are the ones anyone repeating this will make.

Measured across five clean copies, one per id, all running at once:

| id | Enum | Controller | Nodes | Cards |
|---:|---|---|---:|---:|
| 0 | `eUnknownRouter` | Nucleus 2 | 15 | 12 |
| 1 | `eCygnusRouter` | Nucleus 2 | 15 | 12 |
| 2 | `ePyxisRouter` | Nucleus 2 | 15 | 12 |
| 3 | `eSirius800Router` | Nucleus 2 | 15 | 12 |
| 4 | `eVegaRouter` | **Vega Controller** | **6** | **0** |

The node tables for 0–3 are byte-identical. Only 4 differs.

**Mistake one:** we compared 2 against 3 — two identical configurations — and concluded the id changes nothing.

**Mistake two:** we reused a folder that had already run. The zip ships `LocalRouter.dat`, `Persistence_*.dat` and `CentraController.dccp_config`, so a folder keeps the model it wrote and changing the id in place can change nothing observable even when the id does matter. Unpack a fresh copy per id.

Three consequences, each of which cost a wrong conclusion:

1. `CENTRA_SIMULATED_CARDS` does nothing on a Vega — it is a self-contained 2RU router with no card slots.
2. The unit address is stable where the slot number is not. The XY Panel is `0000-81-00` on both models, but slot 14 on a Nucleus and slot 5 on a Vega. Anything keyed on a slot number breaks between models — the reason for #1037.
3. The router *model* — matrices, levels, sizes, salvos, names — comes from persistence rather than from the id, and is identical on all five. That half of the first comparison was right.

## Also recorded

- Several instances run side by side, which is how the table was measured — but **four** ports in `config.xml` have to differ, not one: `RollCall/SharePort`, `RollCall/BridgePort`, `IP/Adapter/Port`, `Web/XPCPort`. Missing one leaves the second instance up and silently crippled.
- The IPShare wedges after roughly three connections in quick succession: accepted at TCP, never answered, cleared only by a restart. A property of the vendor gateway, but it shapes how the emulator tier is driven.

Section numbering fixed — there were two section 9s.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -93,7 +93,7 @@ concerns here.
 
 | Folder | State |
 |---|---|
-| `internal/snell-rollcall/` | **in progress** (epic #1009, gate lifted by the owner 2026-09-07). The ADR-0025 step-1 audit is done and the wire is validated against a live oracle on `10.6.250.105` — consumer read/write/subscribe proven against the real stack, and a throwaway producer spike rendered and driven by RollCall Control Panel 4.12.48. Landed so far: `CLAUDE.md` (wire context) + `docs/` (audit, DM/UI analysis, 1 623 captured frames, 560 menu lines). Still to come, one PR each: `codec/` · `session/` · `consumer/` · `provider/` · `wireshark/` · `integration/` · `testdata/`. No Go code and no registry entry yet. |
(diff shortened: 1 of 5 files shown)


by-rune:

PR #1034, opened 2026-09-07
feat(snell-rollcall): canonical export and per-card device models

by-rune:
Stacked on #1033. ADR-0022 asks for one device model per **card type** rather
than per slot, so a frame of twelve identical cards is walked once. RollCall
has exactly the data for that and none of it was wired up.

## What the consumer can now say about itself

`Canonicalize` / `ExportCanonical` turn a walked menu into a canonical tree with
the nesting the device published — a container is a line whose *step* spans a
subtree — and each line becomes the parameter it is:

| Menu line | Canonical |
|---|---|
| number with a divisor | `real` **with its factor**, so the wire value stays recoverable |
| number without | `integer`, range and step as given |
| checkbox | `boolean` |
| editable string | `string`, range is a length |
| display / list / disabled | read-only |

The round trip holds. The committed frame tree, served by our provider, walked
by our consumer and re-exported:

```
127.0.0.1 [node] read
  slot-0 [node] read
    card1 [node] read
      identity [node] read
        product [string] read
      video [node] read
        gain [real] readWrite  min=-60 max=6 factor=10 fmt='%0.2f dB'
        enable [boolean] readWrite
      status [integer] read  min=0 max=9
```

Same shape, same factor, same access as the tree that went in.

## IdentityProbe keys on the model, not the label

A card's name is user-editable — one is `CAM 1 PROC`, the identical card beside
it is `SPARE`. What identifies a model is the vendor's type id and the
firmware's command set, which is what the specification itself says. So the
identity is `<type>@<major>.<minor>.cs<n>`, sanitised for a path (the vendor's
own names include `4929 AES O/P card`). It costs **one message** — a walk there
would be minutes on a large card, and it runs before anything else wants the
link.

## Two defects in shared CLI code, found by using it

- **`writeCanonicalCapture` dispatched on concrete plugin type** and silently
  wrote nothing for anything not in its list of three. `probel-sw08p` has had a
  `Canonicalize` the CLI has never called. It now asks for the *method*.
- **`extract` renamed the capture file while the recorder still held it open**,
  which fails on Windows: no `tree.json`, and a warning nobody would connect to
  the cause. The recorder is closed before the rename.

`dhs consumer rollcall extract` now produces the full `{meta.json, wire.jsonl,
tree.json}` triple with a fingerprint. Verified for both a card with a nested
menu and one with a flat one; acp1 and the CLI suite unaffected.

Coverage stays at 100% in the consumer.

### DIFF
--- a/cmd/dhs/cmd_extract.go
+++ b/cmd/dhs/cmd_extract.go
@@ -102,6 +102,10 @@ func runExtract(ctx context.Context, args []string) error {
 		return fmt.Errorf("canonical capture: %w", err)
 	}
 
+	// Everything that writes to the capture directory is finished with, so the
+	// recorder can be closed before anything is renamed.
+	cleanup()
+
 	// Rename raw.<transport>.jsonl → wire.jsonl so the fixture layout
 	// matches docs/adr/0020-capture-and-fixture-layout.md (+ tests/fixtures/products/README.md) exactly (the protocol is still
 	// identifiable via meta.json). Best-effort: a missing raw file
--- a/cmd/dhs/cmd_walk.go
+++ b/cmd/dhs/cmd_walk.go
@@ -9,11 +9,11 @@ import (
 	"path/filepath"
 	"strings"
 
-	"dhs/internal/acp1/consumer"
 	"dhs/internal/acp2/consumer"
 	"dhs/internal/consumer"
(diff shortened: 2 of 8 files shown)


by-rune:

PR #1038, opened 2026-09-07
fix(snell-rollcall): name a slot by its node address

### LINKED ISSUE #1037: rollcall: a slot's identity does not carry the node address the runbook documents
`internal/snell-rollcall/docs/runbook.md` §2 says:

> `info` prints each slot's identity, and the address is in the JSON

and shows reading it. `GetSlotInfo` fills `Identity` with `name`, `type`, `type_id`, `version`, `services`, `product`, `category` — and no address.

On a controller that is the one field that matters. A slot number is a position in the device's own enumeration, not a place in a frame, and only the address names the node.

Measured on the vendor Centra emulator, the same build run as two router ids:

| `CENTRA_SIMULATED_ROUTER` | controller | nodes | routing interface |
|---|---|---|---|
| 2 (Pyxis) | `Nucleus 2` | 15 | slot 14, `0000-81-00:010` |
| 4 (Vega) | `Vega Controller` | 6 | slot 5, `0000-81-00:007` |

The routing interface moves between slots. Without the address in the identity, nothing in the JSON says where it went.

Fix: add `address` from the session's peer, which is the node's address and already known — no extra round trip.

by-rune:
Closes #1037. Stacked on #1034.

A slot's identity carried `name`, `type`, `type_id`, `version`, `services`, `product` and `category` — and not the one field `docs/runbook.md` §2 told operators to read.

On a controller a slot number is a position in the device's own enumeration rather than a place in a frame, so it names nothing on its own.

## Measured

Two clean copies of the vendor Centra emulator, same build, on separate ports, differing only by `CENTRA_SIMULATED_ROUTER`:

```
======== id 2 (Pyxis) ========          ======== id 4 (Vega) ========
slot address          type               slot address          type
0    0000-08-00:0FF   Nucleus 2          0    0000-08-00:0FF   Vega Controller
1    0000-11-00:0FF   Router Matrix      1    0000-0A-00:0FF   Vega 2RU
2    0000-12-00:0FF   Router Matrix      2    0000-11-00:0FF   Router Matrix
3..12 0000-41..66     12 iCards          3    0000-12-00:0FF   Router Matrix
13   0000-80-00:0FF   TIELINES           4    0000-80-00:0FF   TIELINES
14   0000-81-00:0FF   XY Panel           5    0000-81-00:0FF   XY Panel
```

The XY Panel is `0000-81-00` on both. Its **slot number** is 14 on one and 5 on the other. A script that keys on the slot breaks between models; one that keys on the address does not. Before this change the JSON offered only the slot.

## No extra cost

The address comes from the session's peer, which is already known — no round trip.

## The session index is dropped

`Peer()` carries the peer's session index (`0000-08-01:031`). That is an artefact of one conversation rather than of the node: it differs on every reconnection, and leaving it in would make two walks of one card look like two device models. Reported as `Peer().Device()`.

## Tests

Package stays at 100% statement coverage. `TestGetSlotInfo` pins the exact address; `TestGetSlotInfoAddressesEachNodeSeparately` pins that two slots do not report one address, which is the whole point of the field.

### DIFF
--- a/internal/snell-rollcall/consumer/consumer_test.go
+++ b/internal/snell-rollcall/consumer/consumer_test.go
@@ -470,6 +470,35 @@ func TestGetSlotInfo(t *testing.T) {
 	if info.Identity["category"] == "" {
 		t.Error("the product category should be reported")
 	}
+
+	// The address is the node itself. A slot number is a position in the
+	// device's own enumeration rather than a place in a frame, so on a
+	// controller it names nothing on its own — the routing interface sits on
+	// slot 14 of one model and slot 5 of another — and only the address says
+	// which node answered.
+	if got := info.Identity["address"]; got != "0000-08-01:0FF" {
+		t.Errorf("address = %q, want the address the enumeration gave slot 1", got)
+	}
+}
+
+// TestGetSlotInfoAddressesEachNodeSeparately covers the reason the address is
+// reported at all: two slots are two nodes, and the identity has to say which.
+func TestGetSlotInfoAddressesEachNodeSeparately(t *testing.T) {
+	h := newHarness(t, nil)
+	ctx := context.Background()
+
+	one, err := h.plugin.GetSlotInfo(ctx, 1)
+	if err != nil {
+		t.Fatalf("GetSlotInfo(1): %v", err)
+	}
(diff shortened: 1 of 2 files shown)


by-rune:

PR #1040, opened 2026-09-07
fix(ansible): settle between connections to a gateway we did not start

### LINKED ISSUE #1039: rollcall integration play wedges the vendor emulator it is written to drive
`ansible/playbooks/snell-rollcall-integration.yml` tier 1 runs `info`, `router`, `route` and optionally a take back to back. Each verb opens its own TCP connection.

Measured against the vendor Centra emulator: roughly the third connection in quick succession is accepted at TCP and then never answered —

```
error: rollcall: handshake with 127.0.0.1:2054: rollcall: reply timeout: GETDEVINFO after 3s
```

— repeatable, and cleared only by restarting the controller. The IPShare wedges under connection churn.

So the play cannot pass against the tier it exists to drive. It is only green today because nobody has run it with `ROLLCALL_SIM_HOST` set against a controller that stayed up long enough.

This is a property of the vendor gateway rather than of the protocol, and a real device is likely no better — the specification's own note is that units "don't time sessions out very well (at all), and then run out of available sessions".

Fix: settle between emulator connections, as a variable so an operator can tune or disable it. Same treatment for the live tier, which has the same shape and a device that matters more.

by-rune:
Closes #1039. Stacked on #1038.

The RollCall integration play's emulator tier ran `info`, `router`, `route` and a take back to back. Every verb opens its own connection, and the vendor IPShare does not survive that.

## Measured

Roughly the third connection in quick succession is accepted at TCP and then never answered:

```
error: rollcall: handshake with 127.0.0.1:2054: rollcall: reply timeout: GETDEVINFO after 3s
```

Repeatable, and cleared only by restarting the controller. **The play could not pass against the tier it exists to drive** — it looked green only because nobody had run it with `ROLLCALL_SIM_HOST` set against a controller that stayed up.

With the settle in place, five consecutive connections five seconds apart all handshake:

```
info    -> device 127.0.0.1:2050, rollcall v3
router  -> slot 14, interface version 13, 4 salvos, 100 devices
route   -> {"destination":1,"level":1,"matrix":1,"routed":false,...}
take    -> connected; refused by Full Control ("source or destination not
           installed") because this fresh plant has none configured —
           a device answer, not a connection failure
```

## Scope

The live tier gets the same treatment. A real unit is no better placed: the vendor's own note is that units "don't time them out very well (at all), and then run out of available sessions".

The loopback tier is deliberately left alone — it drives our own provider, which the play starts and stops itself and which does not wedge.

`ROLLCALL_SETTLE_SECONDS` tunes it; `0` disables it.

## Verification limits

This machine is a Windows box with no Ansible control node, so the play was **not executed** here. What was verified: the YAML parses, and the connection sequence the play performs was reproduced by hand against the vendor emulator with and without the settle, which is the behaviour the change targets. A full `ansible-playbook` run needs a Linux control node from the testbed.

### DIFF
--- a/ansible/playbooks/snell-rollcall-integration.yml
+++ b/ansible/playbooks/snell-rollcall-integration.yml
@@ -37,6 +37,10 @@
 #   # plus a live device, read-only:
 #   ROLLCALL_TEST_HOST=10.6.250.105 \
 #     ansible-playbook -i inventory/hosts.ini playbooks/snell-rollcall-integration.yml
+#
+# Every verb opens its own connection and a RollCall gateway wedges under that
+# churn, so the play settles between connections to a device it did not start.
+# ROLLCALL_SETTLE_SECONDS tunes it; 0 disables it.
 
 - name: Snell RollCall integration (emulator + live device + loopback)
   hosts: localhost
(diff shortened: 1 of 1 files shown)


by-rune:

PR #1041, opened 2026-09-07
fix(snell-rollcall): a router address names the node, not the session

by-rune:
Stacked on #1040.

`dhs consumer rollcall router` printed the address with the peer's session index on it:

```
slot         14 (0000-81-00:010)     <- one run
slot         14 (0000-81-00:026)     <- the next run, same unchanged panel
```

The index is an artefact of one conversation rather than of the node. Two runs against a router nobody touched differ, which makes the output useless to diff and misleading to script against.

Dropped, the same way a slot's identity drops it (#1037).

## Tests

`TestFindRouterProbesEveryNode` now pins that the address carries no session index and still names a node. Package stays at 100% statement coverage.


by-rune:

